PR #9, opened 2023-06-15
Add element displacement plot

stefsmeets:
This PR adds an implementation for the element displacement plots (Figure 1) using pymatgen.


src/__init__.py
```diff
@@ -1,5 +1,5 @@
-from .analyse_md import analyse_md
-
 __all__ = [
```
v1kko, line 1: shouldn't there be import statements here? I can't do 
```python
import GEMDAT
GEMDAT.plot_displacement_per_site(...)
```
right now

src/displacements.py
```diff
@@ -0,0 +1,225 @@
...
+    equilibration_steps = 1250
+
+    if not (path_coords.exists() and path_data.exists()):
+        vasprun = Vasprun(
+            vasp_xml,
+            parse_dos=False,
+            parse_eigen=False,
+            parse_projected_eigen=False,
+            parse_potcar_file=False,
+        )
```
v1kko, line 185: Fine for now, but I think we should do the parsing (and configuration) in a separate file (much like `analyse_md` did). then we can still add a `__main__` that calls the plot through this separate file.
In this way it becomes easier to only parse the data once for an arbitrary number of plots. (even if its cached)

v1kko: But lets fix this later, when we add more plots

v1kko (approved):
Great work, didn't went into the details of all the plots but they look good.

One issue is that I cannot `import GEMDAT.plot_xxx`  or `import GEMDAT.analyze_md` anymore, but we can fix that later

src/__init__.py
stefsmeets: Yep, good point!

PR #11, opened 2023-06-21
Add vibrational properties

stefsmeets:
This PR reads some more variables from the `vasprun.xml` and calculates the vibrational properties.

### DIFF
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -4,10 +4,12 @@ from .displacements import (
     plot_displacement_per_element,
     plot_displacement_per_site,
 )
+from .project import load_project
 
 __all__ = [
     'calculate_displacements',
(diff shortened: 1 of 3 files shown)


stefsmeets:
Hi @v1kko this PR is ready, what do you think?

v1kko (approved):
Looks good, but keep in mind that we have to change the structure around a bit still.
Of course the added plot functions will stay mostly the same so those are great!

The name (probably will) change as well in the future, so you don't have to change it now (unless you are really annoyed by CAPITALS :stuck_out_tongue: ).

stefsmeets:
Cheers, no worries about changing the structure!

And yes, I'm already super annoyed by the CAPITALS. Already too many times I have had `ImportError` 🫣

PR #21, opened 2023-07-03
Adding a Dashboard

v1kko:
Initial version

- Adds a `gemdash` command which runs a dashboard
- Figures added to dashboard

dashboard_1


v1kko:
@stefsmeets Can you try out this dashboard? I am very curious to hear what you think, as you are more of a dashboard person :)

dashboard/run.py
```diff
@@ -0,0 +1,73 @@
+from pathlib import Path
+
+import streamlit as st
+from gemdat import SimulationData, plot_all
+
+#import pygwalker as pyg
+#import streamlit.components.v1 as components
+
```
stefsmeets, line 8:
```suggestion
```

dashboard/run.py
```diff
@@ -0,0 +1,73 @@
...
+#    # Import your data
+#    data_xr = xr.DataArray(extra['speed'],
+#                           dims=['element', 'time']).isel(time=slice(0, 1000))
+#    data_df = data_xr.to_dataframe(name='speed')
+#
+#    # Generate the HTML using Pygwalker
+#    pyg_html = pyg.walk(data_df, return_html=True)
+#
+#    # Embed the HTML into the Streamlit app
+#    components.html(pyg_html, height=1000, scrolling=True)
```
stefsmeets, line 73:
```suggestion
```

pyproject.toml
```diff
@@ -64,8 +68,11 @@ publishing = [
     "build",
 ]
 
+[project.scripts]
+gemdash= "gemdat.dashboard:gemdash"
+
 [tool.setuptools]
-package-dir = {"gemdat" = "src", "pymatgen.analysis.gemdat" = "src"}
+package-dir = {"gemdat" = "src", "pymatgen.analysis.gemdat" = "src", "gemdat.dashboard" = "dashboard" }
```
stefsmeets, line 75: First, I don't think this should be a module. Second, why is this code not in `src/dashboard`?

stefsmeets (approved):
Works great! I love the name gemdash and I think this will make a great demo.

1. Selecting a `vasprun.xml` is a bit cumbersome.
2. I prefer something like `vasprun.xml.cache` (`$FILENAME.cache`) over `cache` as the cache name. I think we should hide the cache name in the dashboard from the user.
3. You could hide the traceback and show a nice message if the data cannot be found:

```python
if not filename:
    st.info('Select simulation data file to continue')
    st.stop()
```

5. The frequency vs. occurance plot is missing the meanfreq value (https://github.com/GEMDAT-repos/GEMDAT/pull/11)

pyproject.toml
v1kko: The benefit of a module here would be that the dashboard is quite clearly separated from the code, this makes it easy to package it separately, which in turn makes it easy for a user to only use gemdat, if they are not interested in gemdash and its dependencies.

But doing it the way of extra dependencies in the package itself is possible as well so the tradeoff here is:
- `pip install gemdat[gemdash]` to enable `gemdash`
Or support:
-  `pip install gemdat` for only gemdat
- ` pip install gemdash` which transparently pulls `gemdat` if not installed

I think I prefer the second option, but probably the first option would be less maintenance :sweat_smile:

v1kko: Ok I will go with the first option for now, makes more sense in this stage

v1kko:
> Works great! I love the name gemdash and I think this will make a great demo.
> 
>     1. Selecting a `vasprun.xml` is a bit cumbersome.
I see that there is a 'file_uploader` which might be able to work with server side files, let me try that
> 
>     2. I prefer something like `vasprun.xml.cache` (`$FILENAME.cache`) over `cache` as the cache name. I think we should hide the cache name in the dashboard from the user.
I agree, this is good to have as a default, not only in the dashboard, we should make this the default
> 
>     3. You could hide the traceback and show a nice message if the data cannot be found:
> 
> 
> ```python
> if not filename:
>     st.info('Select simulation data file to continue')
>     st.stop()
> ```
Yes that's nice

pyproject.toml
stefsmeets: We could have `src/gemdat` and `src/gemdash` for separation as well. 

I think option 2 would require the dashboard to be in a separate repo. I think the benefit of this is that it can be installed separately, but also that we can manage the dependency on `gemdat` better. I.e. updating gemdat does not immediately break gemdash if we pin the version.

stefsmeets:
The `file_uploader` uploads the entire dataset, so it's not great for our 2 GB example though. 

Would this be something:
https://gist.github.com/benlansdell/44000c264d1b373c77497c0ea73f0ef2

PR #29, opened 2023-07-05
Revise extras

stefsmeets:
This PR revises the way extras are handled and passes them around as a `SimpleNamespace`. This makes it a bit easier to pass them around and get the attribrutes. We can convert this to a class/dataclass at a later stage if necessary.

The `calculate_all` now expect `data` + `extras` but they still return dicts with the data which are then added to `extras`.

### DIFF
--- a/src/data.py
+++ b/src/data.py
@@ -107,12 +107,10 @@ class SimulationData:
 
         self._add_shared_variables(extras)
 
-        for obj in (
-                Displacements,
-                Vibration,
-                Tracer,
-        ):
-            extras.__dict__.update(obj.calculate_all(self, extras=extras))
+        extras.__dict__.update(Displacements.calculate_all(self,
+                                                           extras=extras))
+        extras.__dict__.update(Vibration.calculate_all(self, extras=extras))
+        extras.__dict__.update(Tracer.calculate_all(self, extras=extras))
 
         return extras
 
--- a/src/data.py
+++ b/src/data.py
@@ -76,7 +76,7 @@ class SimulationData:
                       z_ion: float = 1.0,
                       n_parts: int = 10,
                       dist_collective: float = 4.5):
-        """Calculate extra parameters and place them in `.extras` attribute and return self.extras.
+        """Calculate extra parameters and return them as a simple namespace.
 
         Parameters
         ----------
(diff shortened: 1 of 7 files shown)


stefsmeets:
Hi @v1kko I revised the extras part in data as discussed today. Do you have any thoughts on this before I merge this PR?

src/data.py
```diff
@@ -68,32 +67,66 @@ def dict(self) -> dict:
...
+            n_parts=n_parts,
+            dist_collective=dist_collective,
+        )
+
+        self._add_shared_variables(extras)
+
+        extras.__dict__.update(Displacements.calculate_all(self,
+                                                           extras=extras))
+        extras.__dict__.update(Vibration.calculate_all(self, extras=extras))
+        extras.__dict__.update(Tracer.calculate_all(self, extras=extras))
```
v1kko, line 113: We could put everything in a dict here, and in the end create a namespace from it, instead of first creating the namespace, and then updating it as a dict

src/data.py
```diff
@@ -68,32 +67,66 @@ def dict(self) -> dict:
             for slotname in self.__slots__  # type: ignore
         }
 
-    def calculate_all(self, **kwargs):
-        """Calculate extra parameters and place them in `.extras` attribute and return self.extras."""
-        self.config = kwargs
+    def calculate_all(self,
```
v1kko, line 70:
```suggestion
   from functools import cache
   @cache
    def calculate_all(self,
```
That might just work now

v1kko (approved):
Nice, it looks already much better.

I have one question left, what is the difference between `data` and `extras` for the compute_all functions, currently it is whether this data is calculated by us or from pymatgen directly.

Is this distinction important for the compute all functions? Or is it better/easier for these functions to only get one object (namespace/dict), or even to access all data through keyword arguments?

Its something to think about, we can always change it later when we know the effect of the sites calculations on the code.

src/data.py
stefsmeets: Yeah, I considered that. I like the namespace because it's easier to type and we can change it for a (data)class at a later stage.

stefsmeets:
> I have one question left, what is the importance of the difference between `data` and `extras` for the compute_all functions, currently this distinction is wether this data is calculated by us or from pymatgen directly.

Exactly, I view `data` as an immutable representation of the simulation results. The `extras` are calculated depending on user input.

src/data.py
stefsmeets: Indeed, but I don't think it will give any performance benefits at this moment.

PR #53, opened 2023-07-12
Fix integration test

stefsmeets:
This PR updates the values in the integration test after the changes in #52. It also fixes a crash because the labels syntax has changed.


stefsmeets:
I'm going to merge this because it looks close enough.

PR #81, opened 2023-07-27
Refactor `SitesData`

### LINKED ISSUE #34: Revise how `SitesData.*_parts` data are being defined
To calculate statistics, we are calculating dedicated *_parts attributes on `SitesData`. These inputs for these analyses are no different than the full data set, just on a smaller subset.

So, instead of having dedicated *parts* variables, like `$VARIABLE_parts`, consider feeding `SitesData` with the parent data already split (probably `atom_sites` or `all_transitions`. Use these as a basis for subsequent calculations.

Basically, instead of:

`parent_data -> SitesData -> derived_data -> split -> derived_data_parts`

One full SitesData instance for entire timeseries with attributes containing lists of parts data.

Do:

```
parent_data -> SitesData -> derived_data
parent_data -> split -> SitesData -> derived_data_parts
```
One SitesData instance for entire timeseries + list of SitesData instances (one for each part).

stefsmeets:
This PR refactors how parts are being calculated on `SitesData`. This avoid the use of custom code for the parts, and makes use of existing methods on `SitesData` where possible.

Closes #34

### Todo

- [x] Can this just be based off `atom_sites`? (avoid calculating `split_transitions`) -> it can't

### DIFF
--- a/src/sites.py
+++ b/src/sites.py
@@ -116,7 +116,7 @@ class SitesData:
 
         These correspond to the axes in `.coll_matrix`.
         """
-        return ['->'.join(key) for key in self.rates]
+        return ['->'.join(key) for key in self.jumps]
 
     def calculate_dist_close(self, data: SimulationData,
                              vibration_amplitude: float):
--- a/src/sites.py
+++ b/src/sites.py
@@ -99,7 +99,7 @@ class SitesData:
 
         # calculate some statistics
         if extras.n_parts > 1:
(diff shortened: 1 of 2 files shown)


stefsmeets:
I think this is good to go, do you have any feedback @v1kko ?

v1kko (approved):
looks good to me,
I think I see some nice opportunity for parallel computation per part, but we can add that later, and make that an issue for now

Issue #62, opened 2023-07-17
Rename `structure` argument to SitesData constructor to avoid confusion with `data.structure`


v1kko:

v1kko:
for example `material`

stefsmeets:
I don't think it's confusing 😅

v1kko:
as an anecdote: I put `data.structure` where `structure` was expected and then I assumed the plot was broken for a week :smiley:

stefsmeets:
Which plot?

Most of the breakage comes from incompatble lattices between `data.structure` and `sites.structure`. The data generated by the matlab code was using the wrong supercell (1,1,1), should be (2,1,1).

v1kko:
`jumps_vs_distance`

stefsmeets:
Ah, so it's more the number of sites than the lattice. This function does not take structure as an argument though.

v1kko:
sorry I got confused, it is calculating the sitesdata for that plot with 
```
sd = SitesData(data.structure)
```
while it should be
```
structure = load_known_material('argyrodite')
sd = SitesData(structure)
```

That gives an out of bounds error in the plot eventually.

PR #85, opened 2023-07-31
Use `Trajectory` in place of `SimulationData`

### LINKED ISSUE #67: Trajectory class
I think we should set up a trajectory class which will make it easier to handle the simulation data.

Most that we care about is the Trajectory anyways. The pymatgen trajectory class is somewhat limited for our use-case.

This could also take some of the methods in the vibration/displacements modules (speed, displacements, etc.)

Just writing some ideas here:

```python
class Trajectory:
   _coords: np.ndarray[$time, $site, $xyz]
   structure: pymatgen.core.Structure

   @property
   def n_steps(self):
       # return number of steps after equilibration time

   @property
   def coords(self):
       return self._coords[self.n_steps:, ...]

   def displacements(self, element: list[str] | str | None=None):
       # Return displacements for all / selected elements

   def speed(self, element: list[str] | str | None=None):
       # Return speed for all / selected elements

   def set_equilibration_time(self, equilibration_time):
       # sets starting point for data

   def get_coords_for_element(self, label):
       # replaces diff_coords, which is somewhat poorly named

```

### Todo

- [x] Add some sort of `Trajectory.metadata` attribute to track some sort of global simulation parameters like temperature
- [x] Replace test trajectories by fixtures
- [ ] What makes sense to move from `calculate_all()` to `GemdatTrajectory`?
- [x] Add intuitive method to `Trajectory` to easily get coordinates for diffusing atom
- [ ] Add `Trajectory.from_PymatgenTrajectory`
- [ ] Update readme.md with new API

stefsmeets:
In this PR I'm exploring `pymatgen.core.trajectory.Trajectory` to see if we can use that instead of `SimulationData`. It seems that everything we need is more or less available from this object alone. 

The goal is to replace `SimulationData` as the data structure, and use `Trajectory` instead. To start, I am making this a subclass to see what we need to do to get this to work.

One of the challenges will be to structure the data in such a way that we can drag along all the `extras`.

Addresses part of #67

Also helps with #62

### Todo

- [x] ~Remove `data.py`~ -> Stores `calculate_all()` function that we still need
- [x] Update tests for `GemdatTrajectory`
- [x] Remove tests for `SimulationData` (do we have any?)
- [x] Update dashboard
- [x] Update plots
- [x] Rename `GemdatTrajectory` -> `Trajectory`, `Trajectory` -> `PymatgenTrajectory`
- [x] Fix rdf calculation / plots, these look off
- [x] Confirm that all plots are the same
- [x] `trajectory_to_volume` takes `Trajectory`, but actually needs `diff_coords`

### Future

- [ ] Add some sort of `Trajectory.metadata` attribute to track some sort of global simulation parameters like temperature
- [ ] Replace test trajectories by fixtures
- [ ] What makes sense to move from `calculate_all()` to `GemdatTrajectory`?
- [ ] Add intuitive method to `Trajectory` to easily get coordinates for diffusing atom

### DIFF
--- a/src/data.py
+++ b/src/data.py
@@ -5,10 +5,10 @@ import numpy as np
 from .calculate.displacements import Displacements
 from .calculate.tracer import Tracer
 from .calculate.vibration import Vibration
-from .trajectory import GemdatTrajectory
+from .trajectory import Trajectory
 
 
-def calculate_all(trajectory: GemdatTrajectory,
+def calculate_all(trajectory: Trajectory,
                   *,
                   diffusing_element: str,
                   known_structure: str | None = None,
@@ -21,7 +21,7 @@ def calculate_all(trajectory: GemdatTrajectory,
 
     Parameters
     ----------
-    trajectory : GemdatTrajectory
+    trajectory : Trajectory
         Input trajectory coordinates
     diffusing_element : str
         Name of the diffusing element
@@ -59,8 +59,7 @@ def calculate_all(trajectory: GemdatTrajectory,
     return extras
 
 
-def _add_shared_variables(trajectory: GemdatTrajectory,
-                          extras: SimpleNamespace):
+def _add_shared_variables(trajectory: Trajectory, extras: SimpleNamespace):
     """Add common shared variables to extras namespace."""
     extras.n_diffusing = sum(
         [e.name == extras.diffusing_element for e in trajectory.species])
--- a/src/data.py
+++ b/src/data.py
@@ -12,7 +12,6 @@ def calculate_all(trajectory: Trajectory,
                   *,
                   diffusing_element: str,
                   known_structure: str | None = None,
-                  equilibration_steps: int = 1250,
                   diffusion_dimensions: int = 3,
                   z_ion: float = 1.0,
                   n_parts: int = 10,
@@ -27,8 +26,6 @@ def calculate_all(trajectory: Trajectory,
         Name of the diffusing element
     structure : str | None
         Path to cif file or name of known structure
-    equilibration_steps : int
-        Number of equilibration steps
     diffusion_dimensions : int
         Number of diffusion dimensions
     z_ion : float
@@ -41,14 +38,11 @@ def calculate_all(trajectory: Trajectory,
     extras = SimpleNamespace(
         diffusing_element=diffusing_element,
         known_structure=known_structure,
-        equilibration_steps=equilibration_steps,
         diffusion_dimensions=diffusion_dimensions,
         z_ion=z_ion,
         n_parts=n_parts,
         dist_collective=dist_collective,
     )
-    trajectory = trajectory[equilibration_steps:]
-
     _add_shared_variables(trajectory, extras)
 
     extras.__dict__.update(
--- a/src/data.py
+++ b/src/data.py
@@ -1,11 +1,6 @@
-import pickle
-from dataclasses import dataclass
-from pathlib import Path
 from types import SimpleNamespace
-from typing import Optional
 
 import numpy as np
-from pymatgen.io import vasp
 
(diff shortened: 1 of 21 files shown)


stefsmeets:
Hi @v1kko , I refactored the entire code base to use a `Trajectory` instead of `SimulationData`. 

I realized that all the attributes we need are mostly contained in the `pymatgen.core.trajectory.Trajectory`. So I subclassed that to have a starting point, we can eventually move stuff upstream once we settle on a design. I'm quite happy with it, as it makes much of our code more internally consistant. I also find the naming much easier to think about conceptually (trajectory vs data).

There are a couple of todo's that I will leave for later.

What do you think?

src/plot.py
```diff
@@ -4,26 +4,23 @@
...
-
-    - plots, a list of plot names, or just a plot name for the plot you want.
-    - Optional arguments which are passed down to the plotting functions.
+    """Display all or a selection of plots.
 
     Parameters
     ----------
-    data : SimulationData
-        data
+    trajectory : Trajectory
```
v1kko, line 18:
```suggestion
    trajectory : Optional[Trajectory]
```

src/data.py
```diff
@@ -1,192 +1,68 @@
...
+
+
+def calculate_all(trajectory: Trajectory,
+                  *,
+                  diffusing_element: str,
+                  known_structure: str | None = None,
+                  diffusion_dimensions: int = 3,
+                  z_ion: float = 1.0,
+                  n_parts: int = 10,
+                  dist_collective: float = 4.5):
```
v1kko, line 18: Should we move this function to the `calculate` folder ? It feels a bit weird now in an otherwise empty `data.py`

src/trajectory.py
```diff
@@ -0,0 +1,110 @@
+import pickle
+from pathlib import Path
+from typing import Optional
+
+from pymatgen.core import Lattice
+from pymatgen.core.trajectory import Trajectory as PymatgenTrajectory
+from pymatgen.io import vasp
+
+
+class Trajectory(PymatgenTrajectory):
```
v1kko, line 10: I noticed that PymatgenTrajectory has no object variables defined inside the class (like a dataclass or a pydantic object), when we merge this back we might introduce that for them as well. Lets create an issue for this.

src/trajectory.py
```diff
@@ -0,0 +1,110 @@
...
+
+        latt = self.lattices[idx]
+        return Lattice(latt)
+
+    def __getitem__(self, frames):
+        """Hack around pymatgen Trajectory limitations."""
+        new = super().__getitem__(frames)
+        if isinstance(new, PymatgenTrajectory):
+            new.__class__ = self.__class__
+            new.temperature = self.temperature
```
v1kko, line 109: shouldn't we create a conversion function here? a `from_PymatgenTrajectory`, that can also keep track of all the changes we introduce.

v1kko (changes requested):
A huge step towards improving the structure. Some small comments but otherwise great work! 

If we can eventually merge the Trajectory class back to pymatgen that would be awesome.

src/data.py
stefsmeets: Yeah, agree that makes sense. I didn't quite know what to do with this function :sweat_smile:

src/trajectory.py
stefsmeets: The changes are really small, and if we start using convesion functions we lose the benefit of subclassing.

v1kko: I don't this this is true? the subclassing stil exists, its just the changes are contained in a function, which actually makes it easier to port afterwards, i think.

stefsmeets: Ah, I see, that makes sense.

Issue #93, opened 2023-08-01
Refactor ported code


v1kko:
As we ported more, we realized that some structures could be better represented. this issue is to track that.

- [x] Refactor all calculate functions into "property" functions, which are calculated on demand
  - [x] Also add a toggle to calculate on creation of the object. 
- [x] move diffusive element displacement code into Trajectories
- [x] Remove calculate folder and put calculations in corresponding files
- [x] Refactor all plots to use the Trajectory / Sites / Jumps classes where possible, so that we do not have to pass each property independently 
- [x] Math ( or TrajectoryStatistics) class for statistics concerning Trajectories
- [x] remove diffusive_element specifiers on `Trajectory` functions, and instead allow a view on trajectory, like:
```python
li_trajectory = trajectory.where(element='Li', equilibration_steps=1250)
```
- [x] Better way to include all plots then using `__all__`
- [x] write `trajectory.precompute()` and `sites.precompute()`
- [x] Fix the RDF functions

stefsmeets:
I have fixed all the issues listed here, and implemented all the fixes from #94 in these PRs:
- #98
- #102 
- #105
- #110
- #112
- #114

PR #94, opened 2023-08-03
Refactor GEMDAT

v1kko:
- [ ] Implement everything from issue #93 
- [x] Fix all the tests except rdf
- [ ]  Fix the RDF tests
- [x] Test the dashboard

Script I am working with:
```python
from pathlib import Path

from gemdat import Trajectory, plot, Vibration
from gemdat.sites import SitesData
from gemdat.io import load_known_material
import matplotlib.pyplot as plt

trajectory = Trajectory.from_vasprun(Path('../short_simulation/vasprun.xml'))
trajectory = trajectory[1250:]
li_trajectory = trajectory.filter('Li')
structure = load_known_material('argyrodite', supercell=(2,1,1))
li_vibration = Vibration(li_trajectory, fs=1.)

sd = SitesData(structure, li_trajectory, li_vibration)
plot(trajectory=li_trajectory,plots=["displacement_per_site"],show=False)
plot(trajectory=trajectory,plots=["displacement_per_element"],show=False)
plot(trajectory=li_trajectory,plots=["displacement_histogram"],show=False)
plot(plots=["frequency_vs_occurence"],vibration=li_vibration, show=False)
plot(plots=["vibrational_amplitudes"],vibration=li_vibration, show=False)
plot(plots=["jumps_vs_distance"],trajectory=li_trajectory, sites=sd, jump_res=1.0, show=False)
plot(plots=["jumps_vs_time"],trajectory=li_trajectory, sites=sd, show=False)
plot(plots=["collective_jumps"],trajectory=li_trajectory, sites=sd, show=False)
plot(plots=["jumps_3d"],trajectory=li_trajectory, sites=sd, show=False)
plt.show()
```


v1kko:
@stefsmeets Curious what you think about this, 
The changes are quite big, but the numbers fortunately stay almost the same.

I removed the calculate all function and used (cached) class functions to calculate arrays. a `precompute` still needs to be implemented that calls each function once, but its not critical.

I especially like the added `parts` and `filter` function for the `Trajectory`. I am not yet sure it is a very good idea to have functional access to all arrays, although it seems to work nicely.

stefsmeets:
Hi @v1kko This PR seems like it just changes about everything in gemdat, and I'm not entirely sure whether I am onboard with all the changes. What exactly is the goal of this refactor?

This PR is also very large, which makes it difficult for me to review it. Just having a glance, I see lots of small issues. If the PR was smaller, it would be easier for me to go through it without losing focus.

Some comments:
- I looked at the Trajectory refactor, and I don't like where the design of the API is headed in this PR. There are a bunch of methods mixed in there that having nothing to do with a Trajectory in my opinion. Things like `tracer_diffusivity` and `mol_per_liter` make no sense there. I thought we wanted to split those out, but instead it feels like we are circling back to what we had in the first place.
- Is this PR 'finished'? It feels a bit rough around the edges to me looking at the docstrings / type hints / comments etc. There are too many things I want to comment on, it would be easier if this PR was split up. 
- `Vibration.meanfreq` would be better as a method. It has nothing to do with `Vibration` in particular, because it mimicks the matlab function: https://nl.mathworks.com/help/signal/ref/meanfreq.html.
- Some of the differences in values in the integration test are concerning. I would expect all test values to stay exactly the same.

v1kko:
Hi @stefsmeets, 

- This PR is not finished, but I wanted to hear your thoughts on it.
- I agree that  the tracer/mol_per_liter part should be factored into its own math/utilities class, something that can be done quite quickly now.
- The difference for the integration test would be most likely because of the `Sites.parts` function, as most (all?) differences are in that area. I expected that a recomputation from the splitted trajectories would yield the same result, but as you say, it doesn't. The idea behind this was that with recomputation the arrays would not have to be split manually.

I can try to split this pull request in smaller ones, but that will take quite some time

stefsmeets:
Yeah, I fell into the same trap last week with `sites.parts`, where I naively thought I could just use the split trajectory directly. I was wrong 😅

That's why I you need to derive everything from `all_transitions` / `atom_sites{_to,_from}` instead.

stefsmeets:
Hi @v1kko I have taken all of the work for this PR and implemented it in several PRs in the past week, refactoring the code where necesarry. This addresses all the points in #93. If you are ok with it, I'm going to close this issue now.

PR #131, opened 2023-08-30
Updating and adding documentation

v1kko:
List of stuff to do:
- [ ] Make sure all examples work
- [x] Add more examples to documentation
- [ ] Notebooks
  - [ ] Simple
  - [ ] CIF File
  - [ ] Automatic site generation 
- [x] Adding / Using your own CIF file

https://gemdat.readthedocs.io/en/documenting


v1kko:
Hi @stefsmeets, I have a question about the autogeneration of sites, I have created this notebook: https://gemdat.readthedocs.io/en/documenting/notebooks/auto/ and I seem to have some trouble with periodic boundaries. However I might be missing something / using it wrong, what do you think?

stefsmeets:
Seems like some of the found sites are too close together. This does not mean you are using it wrong. You could try playing with the parameters, smoothing the density before passing it, or inspecting the sites it found and removing/merging a couple afterwards.

I also see that this error message is a bit poorly formatted (I never got to properly test this fail state):

```python
ValueError: Crystallographic sites are too close together (expected: >0.3922, got: 0.4022 for 
Too close:{site_i.specie.name}({i}) {site_i.frac_coords} - Li(14) [0.36790987 0.9903192  0.29722658]
Too close:{site_i.specie.name}({i}) {site_i.frac_coords} - Li(13) [0.34949495 0.00459184 0.2877551 ]
```

stefsmeets:
Before we merge this, do you think it's feasible to put the notebooks in a separate repo to avoid bloating this one?

v1kko:
Good question, it might be possible, I would have to figure out if mkdocs supports URL's (probably it does), or remote repos

PR #149, opened 2023-10-10
not afraid

alexandrosvasi:
This is amazing

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -1,2 +1,1 @@
 {!../README.md!}
-not afraid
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -1 +1,2 @@
 {!../README.md!}
+not afraid
```
stefsmeets, line 2: I think this will break the code 

```suggestion
```

stefsmeets (changes requested):

v1kko:
This actually does not change anything

PR #160, opened 2023-10-11
Update volume.py

AstyLavrinenko:
New function to plot the density of diffusion atoms using plotly

### DIFF
--- a/src/gemdat/volume.py
+++ b/src/gemdat/volume.py
@@ -355,3 +355,141 @@ def trajectory_to_volume(
         positions=trajectory.positions,
         voxel_mapping=voxel_mapping,
     )
+
+import numpy as np
+import plotly.graph_objects as go
+import plotly.express as px
+from scipy.ndimage import gaussian_filter
+from skimage import measure
+
+def plot_density(lattice, diff_positions, site_positions, site_labels, resolution = 0.1, save_plot = True):
+    """Returns 3D plot of probability distributing of diffusing atom.
+
+    Parameters
+    ----------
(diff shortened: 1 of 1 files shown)


AstyLavrinenko:
I have added the function to plot the density of diffusion atoms using plotly and save as .html. However, I didn't integrate it with the class.

stefsmeets:
Thanks! No worries that it is not integrated, this is something we will take care of. I'll get back to this.

stefsmeets:
I have moved the code to the `plots` submodule. For the isosurface plot I want to use the `Volume` we have as input directly. I think a nice way to go about this is to do the marching cubes on the existing volume (which is in fractional coordinate system). And then convert the vertices to cartesian coordinates. This avoids double work, and I think this will be a bit more robust than the current implementation.

### Todo

- [ ] Add docstrings
- [ ] Add isosurface plot

stefsmeets (approved):
🚀

AstyLavrinenko:
Nice! Thank you!

AstyLavrinenko:
I have a question. If I understand correctly the points are coordinates of Li during simulation now? Because initially the points were coordinates of sites. I probably can load the structure created from .cif to plot points as sites, but then lattice and cart_coord will not match the simulation cell

stefsmeets:
It can be whatever you want 😅

```python
site_stucture = load_known_material('argyrodite')
plots.density(vol, site_structure)
```

Yeah, that the lattices don't match is an issue I can't do much about. You could create a new structure with the simulation lattice and the coordinates from the cif.

Issue #158, opened 2023-10-11
The trajectory objects only have every 10th ionic step in their frame


v9lgraf:
Trajectory.from_vasprun('structures-that_are_already_done_with_MD/vasprun.xml', exception_on_bad_xml=False)

If I execute the above code, the trajectroy has 12378 structures in its frame. 

The actual MD though has 123776 steps (that 10 times more). Somehow the Trajectory.from_structures() constructor by default only includes every 10th step. That can be a good thing because it makes the analysis lighter, but some information may get lost also. 

This needs to be optional. One of the **kwargs in the Trajectory.from_vasprun() should feed into the Trajectory.from_structures() constructor and control how many structures are skipped. 

Again I can provide a vasprun.xml file for you to test, we just have to discuss how I can transfer it to you since we're talking GB size, how was this done with the vasprun.xml file you currently use for testing ?

v9lgraf:
This actually is resolved can be controlled by the adding the ionic_step_skip keyword, got a bit confused with the vasprun.cache file that is created. All clear now

v1kko:
Is this something we should take into account when using the cache?

Perhaps adding the `kwargs` (if present) to the cache filename, so that for each combination of keywords, a cache is generated?

Issue #157, opened 2023-10-11
Trajectory.from_vasprun Constructor


v9lgraf:
Often our MDs are prgrammed to run for say 120 000 steps. But we also indicate a max wall time on the supercomputer. Frequently, the wall time runs out before the maximum steps are reached. When we have e.g. 119700 steps, then we don't bother to restart the MD. 

The problem that arises from this: The vasrpun file is erroneous and cannot be imported simply by 

Trajectory.from_vasprun('structures-that_are_already_done_with_MD/vasprun.xml')

The error you get with this is cryptical.
the following command solves the problem: 

Trajectory.from_vasprun('structures-that_are_already_done_with_MD/vasprun.xml'**, exception_on_bad_xml=False**)

What I would like: I want the error message to suggest to the user to try to add the "exception_on_bad_xml=False" keyword.

I can provide you with an erronous vasprun.xml file if you need for testing.

PR #161, opened 2023-10-12
Added descriptive error on loading incomplete xml file, append kwargs to cache path

v1kko:
This PR does two things:
- It might happen that runs are incomplete, if this happens add a helpful message for hinting that this might be the case, and how to resolve it #157 
- Append the kwargs to the cache path, because using different kwargs will result in different caches, this links to issue #158 
  - Issue: this leads to very long filenames, it might be possible to store this info _inside_ the cache instead, so this comparison can still be done, and the cache can be refreshed on keyword changes.  however this would require some thought, and this solution is much simpler

### DIFF
--- a/src/gemdat/trajectory.py
+++ b/src/gemdat/trajectory.py
@@ -4,6 +4,7 @@ dynamics simulations."""
 from __future__ import annotations
 
 import pickle
+import xml.etree.ElementTree as ET
 from itertools import compress
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


stefsmeets:
Could we use the hash the kwargs (or part of it) instead?

v1kko:
Thats a great idea! yes, a hash would be ideal for this (risk of collisions should be minimal)

PR #176, opened 2023-10-23
Make sure the aspect ratio of the density plot is similar to the volume

v1kko:
I noticed it was hardcoded to 2,1,1 . So I changed it to `lattice.abc` also adjusted the "eye" location to provide a view relative to `lattice.abc` instead of 2,1,1


stefsmeets:
Nice catch! I changed the code to prefer the volume lattice in line with #175 and removed some defaults.

PR #177, opened 2023-10-23
Improving plots

v1kko:
| | |
|---|---|
| newplot | newplot | 
| newplot | newplot |
| newplot | newplot |


v1kko:
newplot 
newplot
newplot

v1kko:
@stefsmeets can you have a look at this? 

I have two questions:

- should I remove the old plotting methods, or keep both?
- Are there more plotting routines which should include `n_parts` based error-analysis?

src/gemdat/plots/_vibration.py
```diff
@@ -91,3 +94,76 @@ def vibrational_amplitudes(*, trajectory: Trajectory) -> plt.Figure:
...
+    single_metrics = SimulationMetrics(trajectory)
+    metrics = [
+        SimulationMetrics(trajectory).amplitudes()
+        for trajectory in trajectories
+    ]
+
+    maxamp = max(map(lambda metric: max(metric), metrics))
+    minamp = min(map(lambda metric: min(metric), metrics))
+    maxamp = max(abs(minamp), maxamp)
+    minamp = -maxamp
```
stefsmeets, line 128: 🤯

Would it be possible to refactor this without lambda and make it easier to read?

stefsmeets, line 128:
```suggestion
    max_amp = max(max(metric) for metric in metrics))
    min_amp = min(min(metric) for metric in metrics))
    
    max_amp = max(abs(min_amp), max_amp)
    min_amp = -max_val
```

src/gemdat/plots/_vibration.py
```diff
@@ -91,3 +94,76 @@ def vibrational_amplitudes(*, trajectory: Trajectory) -> plt.Figure:
...
+    mean = [df[col].mean() for col in df.columns]
+    std = [df[col].std() for col in df.columns]
+
+    df = pd.DataFrame(data=zip(columns, mean, std),
+                      columns=['amplitude', 'count', 'std'])
+
+    if n_parts == 1:
+        fig = px.bar(df, x='amplitude', y='count')
+    else:
+        fig = px.bar(df, x='amplitude', y='count', error_y='std')
```
stefsmeets, line 158:
```suggestion
    error_y = 'std' if n_parts else None
    fig = px.bar(df, x='amplitude', y='count', error_y=error_y)
```

src/gemdat/plots/_vibration.py
```diff
@@ -91,3 +94,76 @@ def vibrational_amplitudes(*, trajectory: Trajectory) -> plt.Figure:
...
+
+    x = np.linspace(minamp, maxamp, 100)
+    y_gauss = stats.norm.pdf(x, 0, single_metrics.vibration_amplitude())
+    fig.add_trace(go.Scatter(x=x, y=y_gauss, name='Fitted Gaussian'))
+
+    fig.update_layout(
+        title='Histogram of vibrational amplitudes with fitted Gaussian',
+        xaxis_title='Amplitude (Angstrom)',
```
stefsmeets, line 166:
```suggestion
        xaxis_title='Amplitude (Ångstrom)',
```

src/gemdat/trajectory.py
```diff
@@ -361,3 +361,22 @@ def filter(self, species: str | Collection[str]) -> Trajectory:
                               lattice=self.get_lattice(),
                               metadata=self.metadata,
                               time_step=self.time_step)
+
+    def split(self, n_parts: int) -> List[Trajectory]:
```
stefsmeets, line 365:
```suggestion
    def split(self, n_parts: int) -> list[Trajectory]:
```

src/gemdat/trajectory.py
```diff
@@ -9,7 +9,7 @@
 import xml.etree.ElementTree as ET
 from itertools import compress
 from pathlib import Path
-from typing import Collection, Optional
+from typing import Collection, List, Optional
```
stefsmeets, line 12: You can use the built-in list type since 3.9.

```suggestion
from typing import Collection, Optional
```

src/gemdat/trajectory.py
```diff
@@ -361,3 +361,22 @@ def filter(self, species: str | Collection[str]) -> Trajectory:
...
+
+        Returns
+        -------
+        List[Trajectory]
+        """
+
+        interval = np.linspace(0, len(self) - 1, n_parts + 1)
+        subtrajectories = [
+            self[int(interval[i]):int(interval[i + 1])] for i in range(n_parts)
+        ]
```
stefsmeets, line 381: Would be nice to use pairwire here: https://docs.python.org/3/library/itertools.html#itertools.pairwise

```suggestion
        interval = np.linspace(0, len(self) - 1, n_parts + 1, dtype=int)
        subtrajectories = [
            self[start:stop] for start, stop in pairwise(interval)
        ]
```

stefsmeets (changes requested):
Looks good! These plots look way nicer than what we had! I left some comments on the code.

Things that come to mind:
- The dashboard seems to have some caching issues where the plots ar not updated if I play with the toggles and `n_parts`.
- I don't see the jumps and vibrational amplitudes ones in the dashboard.
- Most of the new functions do some pandas work to get the data in a usable format. Does it make sense to return directly from Trajectory/Sites/Transitions etc? Maybe consolidating some of the code in the new plotting functions, and to help separate plotting code from data wrangling code. This makes it more testable as well. And users who want to make their own plots can access the data more easily. Relates a bit to https://github.com/GEMDAT-repos/GEMDAT/issues/32
- Avoid having 2 of the same function in the same submodule. Having both `displacement_per_element` and `displacement_per_element2` makes it a bit confusing. We could have 2 submodules `gemdat/plots/matplotlib` and `gemdat/plots/plotly` to separate them, and only expose in our api. So that `from gemdat import plots; plots.displacement_per_element(...)` points to the plotly one if available.
- Matplotlib functions are way simpler, could we keep those around as the reference implemtation and for testing correctness? I can't find any simple workflow for testing plotly plots.

To address your questions:
1. Yes, until we find a way to test the plotly plots.
2. I don't know, they can easily be added later 😅

As a side note, I also think it becomes time to think about refactoring the dashboard, and make separate pages for each group of plots. It becomes a bit cluttered in my view and unfortunately you can't get the tab state, so you must have a checkbox to enable rdf, density, n_parts, etc.

src/gemdat/plots/_jumps.py
```diff
@@ -90,6 +168,63 @@ def jumps_vs_time(*, sites: SitesData, binsize: int = 500) -> plt.Figure:
...
+    std = [df[col].std() for col in df.columns]
+
+    df = pd.DataFrame(data=zip(columns, mean, std),
+                      columns=['time', 'count', 'std'])
+    df['specie'] = sites.floating_specie
+
+    if n_parts > 1:
+        fig = px.bar(df, x='time', y='count', color='specie', error_y='std')
+    else:
+        fig = px.bar(df, x='time', y='count', color='specie')
```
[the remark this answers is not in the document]
v1kko: That causes error bars to show (even if error_y is none) so better to leave it as is

src/gemdat/plots/_jumps.py
```diff
@@ -90,6 +168,63 @@ def jumps_vs_time(*, sites: SitesData, binsize: int = 500) -> plt.Figure:
...
+
+    fig.update_layout(bargap=0.2)
+
+    fig.update_layout(title='Jumps vs. time',
+                      xaxis_title='Time (steps)',
+                      yaxis_title='Number of jumps')
```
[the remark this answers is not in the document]
v1kko: yes

src/gemdat/plots/_vibration.py
v1kko: doesnt work unfortunately (error bars will still show up)

Issue #92, opened 2023-08-01
Calculate transition energy between sites


v1kko:
with the transition energy we can determine the energy threshold which has to be crossed for a successful jump, if this energy is too low we could inform the user that the sites are not well-defined and should probably be merged.

SCiarella:
Can this be connected to #144 ?

v1kko:
Yes I think you are correct, thanks!

PR #178, opened 2023-10-24
pathways branch

### LINKED ISSUE #144: Implement energy landscapes and pathways from simulation
The goal is to derive from the probability density (of e.g. Li+) a free energy landscape and then use that to trace the lowest energy pathways (direction, tortuosity) that lead to percolation (1D, 2D, 3D) etc.

### TODO

- [x] Convert probability density to free energy: `E = - k * T * ln(P)`. Minima in the free energy should coincide with the maxima in the probability density.
- [ ] Calculate connectivity between (meta)stable positions, e.g.: https://dx.doi.org/10.1021/acs.chemmater.0c03893
- [x] Mimick some of the plots in that paper (Figs. 6-8).
- [x] Calculate barriers (i.e. activation energies)

---

softBV does something similar for crystal structures in CIF format:
- https://scripts.iucr.org/cgi-bin/paper?S2052520618015718
- https://www.dmse.nus.edu.sg/asn/softBVCL.html

SCiarella:
fix #144


stefsmeets:
Nice work! Let me know when this is finished, and I'll happily review this code.

Couple of things so far:
- Please put any documenting notebooks [here](https://github.com/GEMDAT-repos/gemdat-notebooks) to avoid ballooning this repo.
- Do we need a custom Dijksta implementation? Why not use a well-tested one from scipy or networkx?
- Can you add some tests for these new functions?

SCiarella:
The branch has been updated and it is ready for review. In particular, the specific comments by @stefsmeets have been addressed in the following way:

> - Please put any documenting notebooks here to avoid ballooning this repo.

[Done](https://github.com/GEMDAT-repos/gemdat-notebooks/blob/main/paths.ipynb)

> - Do we need a custom Dijksta implementation? Why not use a well-tested one from scipy or networkx?

Switched to networkx

> - Can you add some tests for these new functions?

I am not an expert with testing, so I would like to discuss next time about the 'best practices' in this direction

> - What is this direct path for?

It is used to show that it does not work, as mentioned by @tfamprikis in #144. I have now moved `direct_path` from the main gemdat package to `path.ipynb`, since its use is only pedagogical.

src/gemdat/transitions.py
```diff
@@ -401,3 +402,210 @@ def _calculate_occupancy(atom_sites: np.ndarray) -> dict[int, int]:
...
+
+        # Calculate the path cost
+        path_cost = np.sum(path_energy)
+
+        # Remove all the peaks that have been part of the path
+        peaks_mask[counter] = True
+        for mask in range(counter + 1, len(peaks)):
+            if not peaks_mask[mask] and tuple(peaks[mask]) in path:
+                peaks_mask[mask] = True
+                remaining_peaks -= 1
```
v1kko, line 603: Is this correct? Can a peak not be part of a different path with a different starting point and have a total lower energy?

SCiarella: You are right! I have removed this mask.

v1kko (approved):
Looks good to me if you add a test for the perculating pathway

src/gemdat/transitions.py
```diff
@@ -405,3 +406,197 @@ def _calculate_occupancy(atom_sites: np.ndarray) -> dict[int, int]:
...
+                     (-1, 0, -1), (1, 0, -1), (-1, 0, 1), (0, 1, 1),
+                     (0, -1, -1), (0, 1, -1), (0, -1, 1), (1, 1, 1),
+                     (-1, -1, -1), (1, -1, -1), (-1, 1, 1)]
+
+    G = nx.Graph()
+    for i in range(F.shape[0]):
+        for j in range(F.shape[1]):
+            for k in range(F.shape[2]):
+                if F[i, j, k] >= 0 and F[i, j, k] < max_energy_threshold:
+                    G.add_node((i, j, k), energy=F[i, j, k])
```
stefsmeets, line 467: Use ndenumerate: https://numpy.org/doc/stable/reference/generated/numpy.ndenumerate.html

```suggestion

    G = nx.Graph()
    for index, val in np.ndenumerate(F):
        if 0 <= energy < max_energy_threshold:
             G.add_node(index, energy=energy)
```

src/gemdat/transitions.py
```diff
@@ -405,3 +406,197 @@ def _calculate_occupancy(atom_sites: np.ndarray) -> dict[int, int]:
...
+    for i, j, k in G.nodes:
+        for move in movements:
+            neighbor = _wrap_pbc(np.array((i, j, k)) + np.array(move), F.shape)
+            if neighbor in G.nodes:
+                G.add_edge((i, j, k), neighbor, weight=F[neighbor])
```
stefsmeets, line 472:
```suggestion
    for node in G.nodes:
        for move in movements:
            neighbor = _wrap_pbc(node + np.array(move), F.shape)
            if neighbor in G.nodes:
                G.add_edge(node, neighbor, weight=F[neighbor])
```

src/gemdat/transitions.py
```diff
@@ -405,3 +406,197 @@ def _calculate_occupancy(atom_sites: np.ndarray) -> dict[int, int]:
     occupancy = dict(zip(unq, counts))
     occupancy.pop(NOSITE, None)
     return occupancy
+
+
+def _wrap_pbc(coord: np.ndarray, size: tuple) -> tuple:
```
stefsmeets, line 411: I think this should return a numpy array.

src/gemdat/transitions.py
```diff
@@ -405,3 +406,197 @@ def _calculate_occupancy(atom_sites: np.ndarray) -> dict[int, int]:
...
+        Energy along the path
+    """
+
+    optimal_path = nx.shortest_path(F_graph,
+                                    source=start,
+                                    target=end,
+                                    weight='weight')
+    path_energy = [F_graph.nodes[node]['energy'] for node in optimal_path]
+
+    return [optimal_path, path_energy]
```
stefsmeets, line 507: Does this need its own function?

src/gemdat/transitions.py
```diff
@@ -405,3 +406,197 @@ def _calculate_occupancy(atom_sites: np.ndarray) -> dict[int, int]:
...
+        # Calculate the path cost
+        path_cost = np.sum(path_energy)
+
+        if path_cost < total_energy_cost:
+            total_energy_cost = path_cost
+            best_starting_point = starting_point
+            best_perc_path = path
+            best_perc_path_energy = path_energy
+
+    return total_energy_cost, best_starting_point, best_perc_path, best_perc_path_energy
```
stefsmeets, line 602: Try to avoid returning a tuple. It's often easier to return an object, like a namedtuple or dataclass and access these variables via an attribute.

src/gemdat/transitions.py
```diff
@@ -405,3 +406,197 @@ def _calculate_occupancy(atom_sites: np.ndarray) -> dict[int, int]:
...
+
+    return [optimal_path, path_energy]
+
+
+def find_best_perc_path(
+        F: np.ndarray,
+        peaks: np.ndarray,
+        percolate_x=True,
+        percolate_y=False,
+        percolate_z=False) -> typing.Union[float, tuple, tuple, np.ndarray]:
```
stefsmeets, line 515: Type hints are missing.

src/gemdat/transitions.py
```diff
@@ -405,3 +406,197 @@ def _calculate_occupancy(atom_sites: np.ndarray) -> dict[int, int]:
...
+        Coordinates of the starting site where the best percolating path was found
+    best_perc_path: list
+        List of coordinates of the path
+    best_perc_path_energy: list
+        Energy along the path
+    """
+    X_real, Y_real, Z_real = F.shape
+
+    # Find percolation using virtual images along the required dimensions
+    if not (percolate_x + percolate_y + percolate_z):
```
stefsmeets, line 545:
```suggestion
    if not any(percolate_x, percolate_y, percolate_z):
```

src/gemdat/transitions.py
```diff
@@ -405,3 +406,197 @@ def _calculate_occupancy(atom_sites: np.ndarray) -> dict[int, int]:
...
+                              dtype=F.dtype)
+        extended_F[:, :Y_real, :] = F
+        extended_F[:, Y_real:, :] = F
+        F = extended_F
+    if percolate_z:
+        extended_F = np.empty((F.shape[0], F.shape[1], 2 * F.shape[2]),
+                              dtype=F.dtype)
+        extended_F[:, :, :Z_real] = F
+        extended_F[:, :, Z_real:] = F
+        F = extended_F
```
stefsmeets, line 565: I don't understand this code, is this just tiling `F` in 3 directions? Why not use `np.tile`?

src/gemdat/volume.py
```diff
@@ -288,6 +288,26 @@ def to_structure(
...
+    def get_free_energy(
+        self,
+        kBT,
+    ) -> np.ndarray:
+        """Estimate the free energy from volume.
+
+        Parameters
+        ----------
+        kBT : float
+            The temperature of the thermostat
```
stefsmeets, line 300: What is thermostat?

src/gemdat/volume.py
```diff
@@ -288,6 +288,26 @@ def to_structure(
 
         return structure
 
+    def get_free_energy(
+        self,
+        kBT,
```
stefsmeets, line 293:
```suggestion
        temperature: float,
```

src/gemdat/volume.py
```diff
@@ -288,6 +288,26 @@ def to_structure(
...
+            The temperature of the thermostat
+
+        Returns
+        -------
+        free_energy : ndarray
+            Free energy on the voxel grid
+        """
+        prob = self.data / self.data.sum()
+        free_energy = -kBT * np.log(prob)
+        return np.nan_to_num(free_energy)
```
stefsmeets, line 309: Maybe not for this PR, but for the long term we could consider that all methods like this return an instance of `Volume` with some flag that tells that it is a probability or free energy, and have some `.units` attribute to keep track of the units.

src/gemdat/transitions.py
```diff
@@ -405,3 +406,197 @@ def _calculate_occupancy(atom_sites: np.ndarray) -> dict[int, int]:
...
+    total_energy_cost = float('inf')
+    best_starting_point = None
+    best_perc_path = []
+
+    for starting_point in peaks:
+
+        # Get the end point which is a periodic image of the peak
+        end_point = (starting_point[0] + X_real * percolate_x,
+                     starting_point[1] + Y_real * percolate_y,
+                     starting_point[2] + Z_real * percolate_z)
```
stefsmeets, line 581: Define image above loop.

```suggestion
        end_point = starting_point + image
```

stefsmeets (commented):
Hi @SCiarella , Nice work! Just made a pass through the code, some small comments from my side for now. I will have a look at the notebook tomorrow and the PR as a whole! 😃

src/gemdat/volume.py
SCiarella: Agree

src/gemdat/volume.py
SCiarella: It is more understandable if we say 'temperature of the simulation', so I have made the change.

However, the temperature fluctuates along the trajectory and the value that we use is the one that the simulation is trying to impose (the thermostat).

stefsmeets (changes requested):
I just checked the notebook, and I'm wondering if you can think a bit of how you want people to use this code. I find some parts not very obvious and I think it can be streamlined quite a bit.

Feel free to tackle this in a follow-up PR if that makes more sense.

Let me know what you think!

## Free energy

This is the current api in this PR:

```python
from gemdat.transitions import optimal_path, free_energy_graph

diff_volume = trajectory_to_volume(diff_trajectory, resolution=0.3)

F = diff_volume.get_free_energy(kBT=trajectory.metadata['temperature'])

F_graph = free_energy_graph(F, max_energy_threshold=1e7, diagonal=True)

start_point = (49, 26, 10)
end_point = (10, 0, 8)

path, path_energy = optimal_path(
    F_graph,
    start_point,
    end_point,
)

fig1 = plots.path_on_landscape(diff_volume, path, structure)
fig1.show()

fig2 = plots.energy_along_path(energy_path=path_energy)
fig2.show()
```

Playing around a little bit, this is what I would suggest we aim for:
- we can grab the temperature from the trajectory metadata
- `F` is a an instance of `Volume` or a subclass of `Volume` with _energy_ related methods
- no need for any extra gemdat imports
- does not hide the networkx graph, so users can do whatever they want with it


```python
import networkx as nx

diff_volume = trajectory.to_volume(resolution=0.3)

F = diff_volume.to_free_energy()
G = F.to_graph(max_energy_threshold=1e7, diagonal=True)

source = (49, 26, 10)
target = (10, 0, 8)

optimal_path = nx.shortest_path(G,
                                source=source,
                                target=target,
                                weight='weight')

fig1 = plots.path_on_landscape(diff_volume, optimal_path, structure)
fig1.show()

energy = [G.nodes[node]['energy'] for node in optimal_path]

fig2 = plots.energy_along_path(energy_path=energy)
fig2.show()
```

## Percolation

Same goes for the percolation, my suggestion would be for an api that looks something like this:
- no need to import anything = more discoverable
- returning a dataclass of some sort makes it easier to extend the code
- hides the wrapping in a method

```python
peaks = F.find_peaks()
best_path = F.find_best_perc_path(peaks, perc_xyz=(True, False, False))

print(f"Total Energy required: {best_path.total_energy_cost}")
print(f"Starting Point: {best_path.starting_point}")
print(f"Best Path: {best_path.best_path}")
print(f"Best Path Energy: {best_path.best_path_energy}")

fig1 = plots.path_on_landscape(diff_volume, structure, path=best_path)
fig1.show()

fig2 = plots.energy_along_path(path=best_path)
fig2.show()
```

src/gemdat/plots/plotly/_paths.py
```diff
@@ -0,0 +1,122 @@
...
+        List of the coordinates that define the path
+
+    Returns
+    -------
+    fig : matplotlib.figure.Figure
+        Output figure
+    """
+
+    # Create a colormap to visualize the path
+    colormap = plt.get_cmap('viridis')
```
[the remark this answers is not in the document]
SCiarella: ok

src/gemdat/plots/plotly/_paths.py
```diff
@@ -0,0 +1,122 @@
...
+    fig : go.Figure
+        Output as plotly figure
+    """
+
+    fig = density(vol, structure, as_probability=True)
+
+    x_path, y_path, z_path = zip(*path)
+    x_path = np.asarray(x_path) * vol.resolution
+    y_path = np.asarray(y_path) * vol.resolution
+    z_path = np.asarray(z_path) * vol.resolution
```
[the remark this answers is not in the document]
SCiarella: thanks!

src/gemdat/plots/plotly/_paths.py
```diff
@@ -0,0 +1,122 @@
...
+            z=z_path,
+            mode='markers+lines',
+            line=dict(width=3),
+            marker=dict(size=6,
+                        color=[
+                            'blue' if i == 0 or i == len(x_path) - 1 else 'red'
+                            for i in range(len(x_path))
+                        ],
+                        symbol='circle',
+                        opacity=0.9),
```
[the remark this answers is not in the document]
SCiarella: Thanks for the suggestion. Yes, the color is to mark endpoints differently.

src/gemdat/plots/plotly/_density.py
```diff
@@ -158,7 +158,9 @@ def plot_volume(
                       showlegend=False))
 
 
-def density(vol: Volume, structure: Optional[Structure] = None) -> go.Figure:
+def density(vol: Volume,
+            structure: Optional[Structure] = None,
+            as_probability: bool = False) -> go.Figure:
```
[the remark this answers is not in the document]
SCiarella: Created the 'as_probability' method for Volume

src/gemdat/plots/plotly/_paths.py
```diff
@@ -0,0 +1,122 @@
...
+import matplotlib.pyplot as plt
+import numpy as np
+import plotly.graph_objects as go
+
+from gemdat.volume import Structure, Volume
+
+from ._density import density
+
+
+def energy_along_path(*, energy_path: list) -> plt.Figure:
```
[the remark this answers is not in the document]
SCiarella: moved

src/gemdat/plots/plotly/_paths.py
```diff
@@ -0,0 +1,122 @@
...
+    """
+    fig, ax = plt.subplots()
+
+    ax.plot(range(len(energy_path)), energy_path, marker='o', color='r')
+    ax.set(xlabel='Step', ylabel='Energy')
+
+    return fig
+
+
+def path_on_grid(*, path: list) -> plt.Figure:
```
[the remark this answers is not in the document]
SCiarella: moved

src/gemdat/plots/plotly/_paths.py
```diff
@@ -0,0 +1,122 @@
...
+
+    fig, ax = plt.subplots()
+    ax = fig.add_subplot(111, projection='3d')
+
+    path_x, path_y, path_z = zip(*path)
+
+    for i in range(len(path) - 1):
+        ax.plot(path_x[i:i + 2],
+                path_y[i:i + 2],
+                path_z[i:i + 2],
```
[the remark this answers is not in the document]
SCiarella: Yes that is true. I have fixed it.

src/gemdat/transitions.py
SCiarella: Good call! i have just replaced the full block with a single tile call

src/gemdat/transitions.py
SCiarella: nice, thanks

src/gemdat/transitions.py
SCiarella: done, thanks

src/gemdat/transitions.py
SCiarella: yes because it is also useful for other tasks, like issue #92

src/gemdat/transitions.py
SCiarella: added

src/gemdat/transitions.py
SCiarella: done

src/gemdat/transitions.py
```diff
@@ -405,3 +406,197 @@ def _calculate_occupancy(atom_sites: np.ndarray) -> dict[int, int]:
...
+
+def _wrap_pbc(coord: np.ndarray, size: tuple) -> tuple:
+    """Wraps coordinates in periodic boundaries.
+
+    Parameters
+    ----------
+    coord: np.ndarray
+        Coordinates of one or multiple sites
+    size: tuple
+        Boudnary size
```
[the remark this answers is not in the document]
SCiarella: fixed

src/gemdat/transitions.py
SCiarella: To be a node or an edge of a graph object it needs to be hashable, but I can make it a tuple outside and leave this with a more general np.ndarray

src/gemdat/transitions.py
SCiarella: Created a path class

src/gemdat/transitions.py
stefsmeets: The issue is that this function returns a tuple, which means this function does 2 things, which means I must deal with it in my code if I'm only interested in the path. This is poor separation of concerns. It's not obvious to me why it should return the energy as well. If you must return multiple values, it's usually easier in the long term to return an object (see other comment as well), and hide the energy list behind a method or property. Or split in two functions.

Personally, since this function is so trivial, I think it's easier to write it out in the function calling it.

src/gemdat/transitions.py
stefsmeets: Yeah, that makes sense. I saw a bunch of type conversions from array<>tuple, which is what triggered this comment. Either way is okay.

src/gemdat/path.py
```diff
@@ -0,0 +1,211 @@
...
+    """
+
+    # Define possible movements in 3D space
+    movements = [(1, 0, 0), (-1, 0, 0), (0, 1, 0), (0, -1, 0), (0, 0, 1),
+                 (0, 0, -1)]
+    if diagonal:
+        movements.extend([(1, 1, 0), (-1, -1, 0), (1, -1, 0), (-1, 1, 0),
+                          (1, 0, 1), (-1, 0, -1), (1, 0, -1), (-1, 0, 1),
+                          (0, 1, 1), (0, -1, -1), (0, 1, -1), (0, -1, 1),
+                          (1, 1, 1), (-1, -1, -1), (1, -1, -1), (-1, 1, 1)])
```
v1kko, line 94:
```suggestion
        np.append(movements, [(1, 1, 0), (-1, -1, 0), (1, -1, 0), (-1, 1, 0),
                          (1, 0, 1), (-1, 0, -1), (1, 0, -1), (-1, 0, 1),
                          (0, 1, 1), (0, -1, -1), (0, 1, -1), (0, -1, 1),
                          (1, 1, 1), (-1, -1, -1), (1, -1, -1), (-1, 1, 1)])
```

v1kko (commented):
Looks good to me, I suggested a few small changes, but otherwise I think its good to go, any remaining comments can be converted into issues :rocket:

src/gemdat/path.py
```diff
@@ -0,0 +1,211 @@
...
+    size: tuple
+        Boundary size
+
+    Returns:
+    -------
+    wrapped_coord: np.ndarray
+        Coordinates inside the PBC
+    """
+    wrapped_coord = coord % size
+    return wrapped_coord
```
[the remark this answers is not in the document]
SCiarella: Ok, I have removed the function

src/gemdat/path.py
```diff
@@ -0,0 +1,211 @@
...
+
+    Returns
+    -------
+    G : nx.Graph
+        Graph of free energy
+    """
+
+    # Define possible movements in 3D space
+    movements = [(1, 0, 0), (-1, 0, 0), (0, 1, 0), (0, -1, 0), (0, 0, 1),
+                 (0, 0, -1)]
```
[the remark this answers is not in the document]
SCiarella: changed movements to np.array

SCiarella:
I have updated the branch as suggested.  The remaining comments about API and usage can be discussed in another issue

v1kko:
> I have updated the branch as suggested. The remaining comments about API and usage can be discussed in another issue

Perfect, after you create the issue, let's merge this :rocket:

PR #184, opened 2023-11-15
Fix number of colors for plot_points

### LINKED ISSUE #183: Issue with density plot colors for sites
When using the density plot with px.colors.qualitative.G10 as the color scale, some labels don't have assigned colors and are consequently not plotted. The number of colors available in px.colors.qualitative.G10 is less than the number of sites, resulting in only one label having an assigned color in the dictionary.

AstyLavrinenko:
Closes #183


v1kko:
Stef is on vacation, so I had a look instead, and I think it looks good :+1:

AstyLavrinenko:
Thanks Victor!

PR #190, opened 2023-11-20
Fix squared radius in shape_analysis

### LINKED ISSUE #189: Circle plots with squared radius in shape_analysis
In the shape_analysis, there is an issue with the visualization of circles on plots. The problem arises when circles are plotted with squared radius values that, in certain cases, exceed the defined threshold. This leads to incorrect visualization

AstyLavrinenko:
Closes #189


stefsmeets:
Thanks, that makes sense. I will update the test data so that the tests will pass and then this can be merged.

stefsmeets (approved):
:rocket: 

## Todo

- [ ] Update test data
- [x] Fix formatting to make CI happy

PR #194, opened 2023-11-21
Fix mismatch with i variable in activation_energies

### LINKED ISSUE #193: Mismatch with i variable in activation_energies function
self.atom_locations_parts[i][site_start] gives an error. Variable i in atom_locations_parts comes from number of parts, and i comes from number of site_pairs in the function

AstyLavrinenko:
Closes #193

### DIFF
--- a/src/gemdat/sites.py
+++ b/src/gemdat/sites.py
@@ -236,7 +236,8 @@ class SitesData:
 
             part_time = self.trajectory.total_time / n_parts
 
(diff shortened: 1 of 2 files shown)


stefsmeets (approved):
Nice work! I'm not entirely sure why the original code was wrong, but I'm happy to merge these changes. I updated the integration test to match the slightly different results from the fix.

PR #195, opened 2023-11-21
Add mean squared displacement plot

### LINKED ISSUE #192: MSD plot
Add a plot for mean square displacement

AstyLavrinenko:
Closes #192


src/gemdat/plots/plotly/_displacements.py
```diff
@@ -66,6 +66,62 @@ def displacement_per_element(*, trajectory: Trajectory) -> go.Figure:
...
+
+    species = trajectory.species
+
+    for sp, distances in zip(species,
+                             trajectory.distances_from_base_position()):
+        grouped[sp.symbol].append(distances**2)
+
+    for symbol, sq_distances in grouped.items():
+        mean_sq_disp = np.mean(sq_distances, axis=0)
+        std_disp = np.std(sq_distances, axis=0)
```
v1kko, line 95: Calculating the std in this way can result in a negative MSD, which is (unless the displacements can be imaginary) not quite correct.

newplot

I think it might not make much sense to include the std here, but I leave this up to you

v1kko (commented):
Looks great, I have one comment, otherwise I think it can be merged. Also, feel free to add yourself as author to the CITATION.cff file!

PR #222, opened 2023-12-12
Move dashboard to its own repo

### LINKED ISSUE #206: Consider moving the dashboard to its own repo
The dashboard is not a priority for this project. Instead focus on jupyter notebooks. The dashboard can move to its own repository so that we can pin a gemdat version and develop/test it in isolation.

stefsmeets:
I have moved the dashboard to its own repo here: https://github.com/GEMDAT-repos/gemdat-dashboard

I removed the corresponding tests, github actions, dependencies, and updated the documentation. 

Closes #206


stefsmeets:
Hi @v1kko do you have any thoughts or feedback on this PR?

docs/dashboard.md
```diff
@@ -1,17 +1,25 @@
-Gemdash includes a dashboard to interactively select your input data and experimental parameters. This makes it straightforward to explore your data without any coding experience required.
+GEMDAT includes a dashboard to interactively select your input data and experimental parameters. This makes it straightforward to explore your data without any coding experience required.
+
+The dashboard is available from: https://github.com/GEMDAT-repos/gemdat-dashboard
 
 To use the dashboard, first make sure you install all the dependencies:
 
 ```
-pip install gemdat[dashboard]
+pip install -r requirements.txt
```
v1kko, line 8: It would be nice if we can do `pip install gemdash-dashboard` (or `pip install gemdash`) Is there a reason you chose a "local" install method here?

v1kko: It would also mean that we get to keep the `gemdash` command, instead of having to `python gemdash.py`

v1kko (commented):
Looks good, I will have a separate look at the gemdat-dashboard repo. I have only one comment/question. But it should work nicely otherwise

docs/dashboard.md
stefsmeets: I thought about that, and came to the conclusion that I didn't want to go through the hoops of setting up all the boilerplate and upload to pypi etc at this point. Especially because I'm not certain how much effort we want to put into it. But that would definitely be nice to have.

stefsmeets: Yeah, and to keep the `gemdash` command on the dashboard side completely.

PR #225, opened 2023-12-13
Allow custom lattice for density plot and `ShapeAnalyzer`

### LINKED ISSUE #198: Default to lattice from trajectory for analysis
`SitesData` should prefer the lattice from the trajectory instead of model structure. i.e. just load fractional coords from ideal structure.

We could also make it so that the lattices *must* be the same, so that the user needs to take care of this. Alternative, perhaps scale input ideal lattice to trajectory to enforce lattices are the same.

stefsmeets:
This PR makes gemdat a bit more flexible to specify custom lattices.

Closes #198

### DIFF
--- a/src/gemdat/plots/plotly/_density.py
+++ b/src/gemdat/plots/plotly/_density.py
@@ -116,6 +116,7 @@ def plot_volume(
     volume: Volume,
     *,
     fig: go.Figure,
+    lattice: Lattice | None = None,
     colors: list[str] = ['red', 'yellow', 'cyan'],
     isovals: list[float] = [0.25, 0.10, 0.007],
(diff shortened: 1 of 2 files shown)


stefsmeets:
Hi @AstyLavrinenko would this help to reduce lattice matchmaking? 😅 Let me know if there are other areas where this comes up.

AstyLavrinenko:
Hi @stefsmeets 
I've tested it out with my structure, and things seem to be running smoothly. I'm quite pleased with the results – the increased flexibility in customizing the analysis really helps dial down the lattice matchmaking issues. As of now, I haven't come across any other places where lattice mismatch might be a concern. However, if I come across anything, I'll open an additional issue and keep you posted. Thanks! 😄

PR #229, opened 2023-12-19
Switch to bump-my-version

### LINKED ISSUE #227: Use maintained fork for bumpversion
https://github.com/callowayproject/bump-my-version

stefsmeets:
Drop bump2version in favour of bump-my-version.

Closes #227


pyproject.toml
```diff
@@ -101,3 +102,21 @@ required-imports = ["from __future__ import annotations"]
 files = ["src", "tests"]
 allow_redefinition = true
 ignore_missing_imports = true
+
+[tool.bumpversion]
+current_version = "0.9.4"
+
+[[tool.bumpversion.files]]
+filename = "src/gemdat/__init__.py"
+search = "__version__ = '{current_version}'"
```
egpbos, line 111: I think you can leave out the search and replace parts, they should be default. I did a similar PR in the Python template where it works without this, ymmv. https://github.com/NLeSC/python-template/pull/365

stefsmeets: Ah, I did not realize that. It seems to default to a simple string replacement. One issue I ran into before was when my package version matched the CFF version, both were updated.

egpbos: Ah yeah that's tricky... maybe indeed better to keep it explicit then.

Issue #205, opened 2023-11-21
Energy landscape: find path with lowest highest barrier


stefsmeets:
Investigate possibility to find path with lowest highest energy. Is this possible via networkx.

SCiarella:
Connected to #191

SCiarella:
Yes, the `minmax-energy` pathfinding algorithm solves the issue

Issue #191, opened 2023-11-21
More methods for calculating the best pathways


v1kko:
It is a point of discussion which one is correct, however we can implement multiple

- Minimal maximum energy
- Dijkstra but with exponentially scaled energy

stefsmeets:
Closed by #239

PR #239, opened 2024-01-17
Add multiple pathfinding options

SCiarella:
This branch addresses #205 and #191 . 

In addition to Dijkstra algorithm (implemented in main), this branch introduces the following pathfinding algorithms:
* `bellman-ford` common alternative to Dijkstra for negative or cyclic graphs 
* `dijkstra-exp` variation of the Dijkstra algorithm that uses exponential weights
* `minmax-energy` finds the path with the lowest maximum-energy

Notice that all the algorithms are based on [`networkx.shortest_path`](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.shortest_paths.generic.shortest_path.html#networkx.algorithms.shortest_paths.generic.shortest_path) 

 The following figure compares the different algorithms:
output

### DIFF
--- a/src/gemdat/path.py
+++ b/src/gemdat/path.py
@@ -192,14 +192,14 @@ def free_energy_graph(F: np.ndarray,
             if neighbor in G.nodes:
                 exp_n_energy = np.exp(F[neighbor])
                 if exp_n_energy < max_energy_threshold:
-                    weight_exp=exp_n_energy
+                    weight_exp = exp_n_energy
                 else:
-                    weight_exp=max_energy_threshold
-                
+                    weight_exp = max_energy_threshold
+
                 G.add_edge(node,
-                               neighbor,
-                               weight=F[neighbor],
-                               weight_exp=weight_exp)
+                           neighbor,
+                           weight=F[neighbor],
+                           weight_exp=weight_exp)
 
     return G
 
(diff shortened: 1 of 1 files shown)


src/gemdat/path.py
```diff
@@ -190,18 +190,32 @@ def free_energy_graph(F: np.ndarray,
...
+                if exp_n_energy < max_energy_threshold:
+                    G.add_edge(node,
+                               neighbor,
+                               weight=F[neighbor],
+                               weight_exp=exp_n_energy)
+                else:
+                    G.add_edge(node,
+                               neighbor,
+                               weight=F[neighbor],
+                               weight_exp=max_energy_threshold)
```
stefsmeets, line 203:
```suggestion
                exp_n_energy = np.exp(F[neighbor])
                if exp_n_energy < max_energy_threshold:
                    weight_exp=exp_n_energy
                else:
                    weight_exp=max_energy_threshold
                
                G.add_edge(node,
                               neighbor,
                               weight=F[neighbor],
                               weight_exp=weight_exp)
```

src/gemdat/path.py
```diff
@@ -211,22 +225,64 @@ def optimal_path(
...
-    path_energy = [F_graph.nodes[node]['energy'] for node in optimal_path]
+    optimal_path = nx.shortest_path(
+        F_graph,
+        source=start,
+        target=end,
+        weight='weight_exp' if method == 'dijkstra-exp' else 'weight',
+        method='dijkstra'
+        if method == 'dijkstra-exp' or method == 'minmax-energy' else method)
+
+    if method == 'dijkstra' or method == 'bellman-ford' or method == 'dijkstra-exp':
```
stefsmeets, line 249:
```suggestion
    if method in ('dijkstra', 'bellman-ford', 'dijkstra-exp'):
```

src/gemdat/path.py
```diff
@@ -211,22 +225,64 @@ def optimal_path(
...
-                                    target=end,
-                                    weight='weight')
-    path_energy = [F_graph.nodes[node]['energy'] for node in optimal_path]
+    optimal_path = nx.shortest_path(
+        F_graph,
+        source=start,
+        target=end,
+        weight='weight_exp' if method == 'dijkstra-exp' else 'weight',
+        method='dijkstra'
+        if method == 'dijkstra-exp' or method == 'minmax-energy' else method)
```
stefsmeets, line 247:
```suggestion
        if method in ('dijkstra-exp', 'minmax-energy') else method)
```

src/gemdat/path.py
```diff
@@ -211,22 +225,64 @@ def optimal_path(
...
+                source=start,
+                target=end,
+                weight='weight',
+            )
+            minmax_energy = max(
+                [F_graph.nodes[node]['energy'] for node in pruned_path])
+
+            if minmax_energy < max_energy:
+                optimal_path = pruned_path
+                max_energy = minmax_energy
```
stefsmeets, line 278: If possible, consider refactoring this to get the minmax-energy code 'out of the way'. This will also make it easier to unit test this bit.

You could then re-use the code to construct Pathway in one go I believe.

```suggestion
    if method == 'minmax-energy':
        optimal_path = _optimal_path_minmax_energy(...)
    else:
        raise ValueError(f'Unknown method {method}')
        
    path_energy = [F_graph.nodes[node]['energy'] for node in optimal_path]
    path = Pathway(sites=optimal_path, energy=path_energy)
    return path
```

tests/integration/path_test.py
```diff
@@ -5,6 +5,41 @@
 import pytest
 
 from gemdat.io import load_known_material
+from gemdat.path import optimal_path
+
+
+@pytest.vaspxml_available  # type: ignore
+def test_optimal_path(vasp_full_vol, vasp_F_graph):
```
stefsmeets, line 12: You could simplify this code by parametrizing it: https://docs.pytest.org/en/7.1.x/example/parametrize.html#different-options-for-test-ids

This will make it easier to add new cases later on, and to identify which part failed.

Not necessary for this PR, just in case you are interested. For example (different project): https://github.com/duqtools/duqtools/blob/5d6bde0dd1506e0ce570e3dd3c37988ea52ca036/tests/ids/test_ids_operation.py#L147

stefsmeets (approved):
Cool, nice work! Some small suggestions on how to re-arrange the code a bit from my side.

tests/integration/path_test.py
SCiarella: Thanks! I will leave this change out of the PR for now, but I am definitely interested in learning about parametrization

src/gemdat/path.py
SCiarella: Thanks, I have implemented this suggestion. If you agree with the last commit, I think we can merge this branch.

PR #242, opened 2024-01-23
Add method to extend trajectory

### LINKED ISSUE #236: Merge consecutive trajectories
Implement a method for `Trajectory` to load a 'continuation' from a different .XML file and append it to the original trajectory.

stefsmeets:
This PR adds a method to extend the current trajectory with another trajectory. I mimicked the api from `list.extend`. Extending fails if any of the metadata (time step, lattice, etc) does not match. 

Closes #236


stefsmeets:
Hi @SCiarella , Could you review this PR and let me know what you think?

SCiarella:
I found [this](https://pymatgen.org/pymatgen.core.html#pymatgen.core.trajectory.Trajectory.extend) `extend` method from `pymatgen`, which is very similar to what you implemented except for being a bit more strict on the metadata. Can we directly use this one?

stefsmeets:
Ah, I should have checked. Yes that's way better! I will check if it works and then close this PR. Thanks!

stefsmeets:
We may still need to have some custom code to make sure the extended trajectory is actually a gemdat trajectory.

stefsmeets:
OK, seems like the pymatgen implementation covers our use case. I added a test so that we can keep track of this functionality.

SCiarella (commented):
Perfect!

Issue #237, opened 2024-01-10
Use multiple starting point to compute the MSD


SCiarella:
While calculating the MSD as $$MSD(t-t_0) = \langle |\vec{x}(t)-\vec{x}(t_0)|^2\rangle ,$$ it is possible to obtain better statistics by averaging over different initial values $t_0$.

stefsmeets:
Could you point to some instances where this is relevant for GEMDAT?

SCiarella:
The MSD is calculated in the following situations:
* `_displacements.msd_per_element` to plot a curve representing the MSD
* `_shape.shape` to report the average MSD in the plot title

It is possible to obtain a statistically improved estimation (i.e. a smoother curve or a more representative average) by following this suggestion.

stefsmeets:
I had a look at this and I'm not quite sure how we would implement this. Would you like to have a go at it?

SCiarella:
Yes, sure!
This multiple starting point will also be implemented for the rotations in #220 as discussed with @tfamprikis

PR #248, opened 2024-01-25
Improve MSD calculation

SCiarella:
This branch improves the statistics of the MSD calculation, expanding on the idea of the multiple starting points, discussed in #237.

In the limit of infinitely many starting-points, the MSD algorithm can consider all the non-overlapping continuous subtrajectories of length $t_i$ in order to compute all the corresponding $MSD(t_i)$.
Numerically this can be efficiently implemented following the algorithm introduced in [this paper](https://doi.org/10.1051/sfn/201112010), which is well explained in [this stackoverflow thread](https://stackoverflow.com/questions/34222272/computing-mean-square-displacement-using-python-and-fft).

This new MDS-algorithm uses Fast Fourier Transform, and produces estimation which are significantly less sensitive to noise.
These are some plot to showcase the improved performance:
li

It can also be used for multiple species at the same time:
many

And also using plotly:
newplot


src/gemdat/trajectory.py
```diff
@@ -401,3 +428,51 @@ def split(self,
             ]
 
         return subtrajectories
+
+    @property
+    def MSD(self) -> np.ndarray:
```
stefsmeets, line 433: Just a nitpick, it's convention to only use capitals for global values. I also think it's clearer what this does if you spell out and turn it into a function.

```suggestion
    def mean_square_displacement(self) -> np.ndarray:
```

src/gemdat/plots/matplotlib/_displacements.py
```diff
@@ -81,19 +81,16 @@ def msd_per_element(*, trajectory: Trajectory) -> plt.Figure:
...
     fig, ax = plt.subplots()
 
-    for symbol, sq_distances in grouped.items():
-        mean_sq_disp = np.mean(sq_distances, axis=0)
-        ax.plot(mean_sq_disp, lw=0.3, label=symbol)
+    for sp in species:
+        if len(species) > 1:
+            traj = trajectory.filter(sp.symbol)
+        else:
+            traj = trajectory
```
stefsmeets, line 92: I don't understand what this bit does. It loops over the species, but if the lengh of species is bigger then one it filters?

Isn't this essentially the same as this:

```suggestion
    for sp in species:
        traj = trajectory.filter(sp.symbol)
```

src/gemdat/plots/plotly/_displacements.py
```diff
@@ -82,22 +82,19 @@ def msd_per_element(*, trajectory: Trajectory) -> go.Figure:
...
-                             trajectory.distances_from_base_position()):
-        grouped[sp.symbol].append(distances**2)
-
-    for symbol, sq_distances in grouped.items():
-        mean_sq_disp = np.mean(sq_distances, axis=0)
+    for sp in species:
+        if len(species) > 1:
+            traj = trajectory.filter(sp.symbol)
+        else:
+            traj = trajectory
```
stefsmeets, line 91: See comment above.

stefsmeets (approved):
Nice work! I don't entirely understand how the the new msd works, but the plots look nice 😅

A few minor nitpicks from my side.

src/gemdat/trajectory.py
SCiarella: Makes sense! Thanks

src/gemdat/plots/matplotlib/_displacements.py
SCiarella: I was scared about filtering a trajectory multiple times, but it makes no difference at all, you are right!

PR #251, opened 2024-01-31
Refactor sites methods/attrs into other classes

### LINKED ISSUE #240: SitesData contains data and functions that do not belong there
the sitesdata does a lot of extra things that should be separate:

- [x] trajectory
- [x] diff_trajectory
- [x] metrics
- [x] all occupancy methods

These do not belong there, and might be moved, or just removed and computed directly

stefsmeets:
This PR removes the SitesData class and moves all its methods/attributes to other classes.

Closes #240

### Todo

- [x] radial distribution function
- [x] legacy
- [x] update docs
- [x] fix tests
- [x] check if diff_trajectory is passed correctly in transitions -> it seems that jumps.n_floating returns 104 (all species) instead of 48 (Li only). This does not seem right. -> #252 
- [x] is Transitions.floating_species necessary?
- [x] Rename/refactor tests/integration/sites_test.py
- [x] consider renaming `structure` -> `sites` on Jumps/Transitions now that name clash with SitesData is no longer an issue


stefsmeets:
@v1kko Could you review this one?

src/gemdat/transitions.py
```diff
@@ -41,43 +42,63 @@ def __init__(
...
+        """Return number of sites."""
+        return len(self.sites)
+
     @classmethod
     def from_trajectory(
         cls,
         *,
         trajectory: Trajectory,
-        structure: Structure,
+        sites: Structure,
```
v1kko, line 101: why this change from structure -> sites, while the class is still called Structure, that is a bit confusing to me ...

src/gemdat/transitions.py
```diff
@@ -193,7 +194,7 @@ def occupancy(self) -> Structure:
         structure : Structure
             Structure with occupancies set on the sites.
```
v1kko, line 195:
```suggestion
        sites : Structure
            Structure with occupancies set on the sites.
```

tests/integration/jumps_test.py
```diff
@@ -1,39 +1,108 @@
...
+    def test_activation_energies(self, vasp_jumps):
+        activation_energies = vasp_jumps.activation_energies(n_parts=10)
+
+        assert isinstance(activation_energies, pd.DataFrame)
+        assert len(activation_energies) == 1
+
+        row = activation_energies.loc[('48h', '48h')]
+
+        assert isclose(row['energy'], 0.134486, abs_tol=1e-4)
+        assert isclose(row['std'], 0.00405952, abs_tol=1e-6)
```
v1kko, line 61: The values have changed but I remember that you mentioned this and that it is now the old value again (so I leave a comment here, but I assume it is correct now)

v1kko (approved):
So many movements and name changes, but overall everything seems to be in order, I left some small comments, but otherwise I think it is good to go

tests/integration/jumps_test.py
stefsmeets: This was from a bug in #228, see https://github.com/GEMDAT-repos/GEMDAT/pull/253

src/gemdat/transitions.py
stefsmeets: I wanted to make clear that this is not an arbitrary structure, but the collection of known sites that the user inputs. The goal is to make it easier to track what this is. In many places we already used `sites` (previously an instance of SitesData), which now refers to a Structure. So there it is maybe a bit less confusing.

stefsmeets:
There is a remaining issue that the rdf is not calculated correctly. It needs the full trajectory, when passing transitions, it just calculates RDF for 'Li'.

PR #255, opened 2024-02-06
Fix RDF test

### LINKED ISSUE #241: b65577 breaks rdf plot 1, investigate
the results for rdfs have changed, but is not clear why. This might be a good oppurtunity to have a look to refactor it and make it better

stefsmeets:
This PR fixes the rdf test. The issue was that transitions were generated with all time steps, but sites with a subset. This created a mismatch that this patch resolves.

Closes #241


v1kko:
What a catch! :clap:

Issue #204, opened 2023-11-21
Energy landscape: Multiple short paths


stefsmeets:
Between two points there may sometimes be multiple paths with very similar energy. GEMDAT should provide a way (can also be networkx) to obtain and plot these paths, both energy and in 3D.

v1kko:
Depends on #213

PR #258, opened 2024-02-08
Multiple paths

SCiarella:
**This PR has the following two main objectives**

(1) Fix a small bug in the definition of the free energy graph:

Instead of weighting the graph edge between `i` and `j` as
`weight = F[j]`
we now do instead 
`weight = 0.5*(F[i] + F[j])`.
This guarantees that the graph is well balanced and the optimal path is calculated corrected in all edge cases.
The tests have been update according to this new definition.

(2) Address #204 

I have added the function `paths.multiple_paths` that returns a list of `Pathway` object, corresponding to the desired number of optimal paths between two selected sites.
The multiple paths are collected according to [this algorithm](https://doi.org/10.1287/mnsc.17.11.712) (implemented in [networkx](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.simple_paths.shortest_simple_paths.html#shortest-simple-paths) ) , while discarding all the ones that are too similar.
I have updated `energy_along_path` and `path_on_landscape` to accept lists of paths as input and produce plots like this one:


SCiarella:
I have added [this notebook](https://github.com/GEMDAT-repos/gemdat-notebooks/blob/main/multiple_paths.ipynb) to showcase the multiple path functionality.

src/gemdat/path.py
```diff
@@ -187,15 +187,15 @@ def free_energy_graph(F: np.ndarray,
         for move in movements:
             neighbor = tuple((node + move) % F.shape)
             if neighbor in G.nodes:
-                exp_n_energy = np.exp(F[neighbor])
+                exp_n_energy = np.exp(0.5 * (F[node] + F[neighbor]))
```
stefsmeets, line 190:
```suggestion
                weight = 0.5 * (F[node] + F[neighbor])
                exp_n_energy = np.exp(weight)
```

src/gemdat/path.py
```diff
@@ -205,6 +205,101 @@ def free_energy_graph(F: np.ndarray,
...
+    Returns
+    -------
+    difference : float
+        Difference between the two paths
+    """
+
+    # Find the shortest and longest paths
+    shortest_path = path_sites1 if len(path_sites1) <= len(
+        path_sites2) else path_sites2
+    longest_path = path_sites2 if shortest_path is path_sites1 else path_sites1
```
stefsmeets, line 228: There must be a cleverer way to do this, but just keep it simple for readability 😅

```suggestion
    if len(path1) <= len(path2):
        shortest, longest = path1, path2
    else:
        shortest, longest = path2, path1
```

src/gemdat/path.py
```diff
@@ -205,6 +205,101 @@ def free_energy_graph(F: np.ndarray,
...
+    return 1 - (shared_nodes / len(shortest_path))
+
+
+def multiple_paths(
+    F_graph: nx.Graph,
+    start: tuple,
+    end: tuple,
+    method: _PATHFINDING_METHODS = 'dijkstra',
+    Npaths: int = 3,
+    min_diff: float = 0.15,
```
stefsmeets, line 245: Best to force keyword arguments for such functions. This prevents errors when the api changes or if you mix up start/stop at some point.

Another point is to choose good 'opposites' as variable names. Either start<->stop, or begin<->end.

Finally, try to avoid capitals in variable names if you can.

```suggestion
    F_graph: nx.Graph,
    *,
    start: tuple,
    stop: tuple,
    method: _PATHFINDING_METHODS = 'dijkstra',
    n_paths: int = 3,
    min_diff: float = 0.15,
```

stefsmeets: Also can you update the types here for start/stop?

src/gemdat/path.py
```diff
@@ -205,6 +205,101 @@ def free_energy_graph(F: np.ndarray,
...
+        Method used to calculate the shortest path. Options are:
+        - 'dijkstra': Dijkstra's algorithm
+        - 'bellman-ford': Bellman-Ford algorithm
+        - 'minmax-energy': Minmax energy algorithm
+        - 'dijkstra-exp': Dijkstra's algorithm with exponential weights
+    Npaths : int
+        Number of paths to be calculated
+    min_diff : float
+        Minimum difference between the paths
+    """
```
stefsmeets, line 269: Return statement is missing in the docstring

src/gemdat/path.py
```diff
@@ -205,6 +205,101 @@ def free_energy_graph(F: np.ndarray,
...
+                    raise ValueError('Path too similar to another')
+        except ValueError:
+            continue
+
+        list_of_paths_sites.append(path)
+        path_energy = [F_graph.nodes[node]['energy'] for node in path]
+        list_of_paths.append(Pathway(sites=path, energy=path_energy))
+
+        if len(list_of_paths) == Npaths:
+            break
```
stefsmeets, line 298: I think it is not necessary to keep track of `list_of_paths_sites` here, these are already included in `list_of_paths`.

You can also try to refactor that try/except block into its own function, to make the code a bit nicer to read. Breaking out of 2 loops at once is a bit awkward, and refactoring the inner loop into its own function and using return is a common pattern to do it. 

```suggestion
    list_of_paths = [best_path]

    # Compute the iterator over all the short paths
    all_paths = nx.shortest_simple_paths(F_graph,
                                         source=start,
                                         target=end,
                                         weight='weight')

    # Attempt to find the Np shortest paths
    for idx, path in enumerate(all_paths):
       if path_too_similar(path, list_of_paths):
           continue

        path_energy = [F_graph.nodes[node]['energy'] for node in path]
        list_of_paths.append(Pathway(sites=path, energy=path_energy))

        if len(list_of_paths) == Npaths:
            break
```

src/gemdat/plots/matplotlib/_paths.py
```diff
@@ -7,28 +7,42 @@
 from gemdat.volume import Volume
 
 
-def energy_along_path(*, path: Pathway, structure: Structure,
+def energy_along_path(*, paths: Pathway | list[Pathway], structure: Structure,
```
stefsmeets, line 10: Could be a clever use for *paths here. Not sure, just a thought.

```suggestion
def energy_along_path(*paths: Pathway | list[Pathway], *, structure: Structure,
```

src/gemdat/plots/matplotlib/_paths.py
```diff
@@ -77,6 +91,16 @@ def energy_along_path(*, path: Pathway, structure: Structure,
                           rotation=45)
     ax_up.get_yaxis().set_visible(False)
 
+    # If available, plot the other pathways
+    if isinstance(paths, list):
+        for idx, path in enumerate(paths[1:]):
+            if path.energy is None:
+                raise ValueError('Pathway does not contain energy data')
+            ax.plot(range(len(path.energy)),
+                    path.energy,
```
stefsmeets, line 100: Isn't it implied? If you leave out the range, does matplotlib not automatically generate the range?

```suggestion
            ax.plot(path.energy,
```

stefsmeets (approved):
Cool, nice work! Some python suggestions/nitpicks from my side.

src/gemdat/path.py
SCiarella: done!

src/gemdat/path.py
SCiarella: thanks! i missed that!

src/gemdat/plots/matplotlib/_paths.py
SCiarella: Not sure about this, I think if I prefer *, paths for clarity. Is there a technical advantage that you see?

src/gemdat/plots/matplotlib/_paths.py
SCiarella: true!

src/gemdat/path.py
```diff
@@ -205,10 +206,130 @@ def free_energy_graph(F: np.ndarray,
...
+    and its implementation in NetworkX. Only paths that are different by at least min_diff are considered.
+
+    Parameters
+    ----------
+    F_graph : nx.Graph
+        Graph of the free energy
+    start : tuple
+        Coordinates of the starting point
+    stop: tuple
+        Coordinates of the stoping point
```
stefsmeets, line 286:
```suggestion
        Coordinates of the stopping point
```

src/gemdat/plots/matplotlib/_paths.py
stefsmeets: The technical advantage is that with `*paths`, the paths variable is always a list, so you don't need to do any instance checking. It's not super common though so it's a bit awkward to use sometimes. I also think `*, paths` is clearer if documented well.

src/gemdat/path.py
```diff
@@ -205,10 +206,130 @@ def free_energy_graph(F: np.ndarray,
...
+    -------
+    difference : float
+        Difference between the two paths
+    """
+
+    # Find the shortest and longest paths
+    if len(path1) <= len(path2):
+        shortest, longest = path1, path2
+    else:
+        shortest, longest = path2, path1
```
stefsmeets, line 230: Cool idea with the min/max! I think this should work:

```suggestion
    shortest, longest = sorted((path1,path2), key=len)
```

stefsmeets (approved):
Looks good to me! 🚀

src/gemdat/path.py
SCiarella: This looks neat!

Issue #286, opened 2024-03-18
rectilinear plot


tfamprikis:
should take a cartesian directory array (e.g. created by `Orientations.get_unit_vectors_trajectory()`) argument alternatively to (instead of?) an `Orientations` object. 

also, right now assumes that a primitive to conventional conversion needs to happen but this should be optional and not default.

stefsmeets:
@SCiarella is this solved by #293?

SCiarella:
I would say that this specific problem is solved.
We will open a new issue if #293 is not general enough.

PR #293, opened 2024-04-03
Fix Orientations IO

### LINKED ISSUE #287: Orientations methods wrongly assume primitive-to-conventional transformation
e.g.

`Orientations.get_symmetric_traj()` first does `Orientations.get_conventional_coordinates()`

these should be independent transformations. sometimes you wanna use one, the other, both, etc.

### LINKED ISSUE #288: orientations.get_conventional_coordinates()
should rewrite to a general `get_transformed_coordinates()` that takes as argument a 3x3 transformation matrix.

the current transformation matrix hard written in `orientations.get_conventional_coordinates()` can be a default since its a usual case.

### LINKED ISSUE #289: orientations function to return non-normalized cartesian vectors
i.e. something like `get_unit_vectors_trajectory()` that forgoes the normalization to a unit vector, i.e. retains the information of bond/vector lengths.

rn the only way to do this is to go through `get_conventional_coordinates()` which has the side effect of applying the primitive to conventional transformation that is not always needed.

this function should then be used in all the `Orientations` transformation functions (e.g. `get_symmetric_trajectory()`) and plots (`rectilinear_plot`, `bond_length_distribution`)

SCiarella:
This PR improves the `Orientations` class.
In particular, this PR implements a **Strategy** design pattern to allow users to select which transformations to use and their order.

Previous methods of `Orientations` have been reformatted into the abstract classes `Normalize`, `Conventional`, and `Symmetrize` that perform their respective transformation. The user will then access `Orientation.transformed_trajectory` to get the result of any series of transformations, instead of using different `_get*` methods.

 The specific outcomes of the PR are the following:
* closes #287 by granting control about which transformation to use and their order. This is done via `set_transformation` and `execute_transformation`
* closes #288 by allowing different primitive cells using the setter method `Orientations.prim_to_conv_matrix`
* closes #289 by allowing the user to exclude or include the `Normalize` transformation as needed.

The corresponding [notebook](https://github.com/GEMDAT-repos/gemdat-notebooks/blob/main/rotations.ipynb) has been updated.


stefsmeets (commented):
Hi @SCiarella , interesting solution to this problem 😅

I'm wondering if we can simplify it a bit, from:

```python
transforms = ['Normalize', 'Conventional', 'Symmetrize']
vasp_orientations.set_transformation(transforms)
transformed_traj = vasp_orientations.execute_transformations()
```

to

```python
transforms = ['Normalize', 'Conventional', 'Symmetrize']
transformed_traj = vasp_orientations.transform(transforms)
```

or return `Orientations` with modified `.unit_vec_trajectory` so that the operations can be chained:

```python
transformed = vasp_orientations.normalize().conventional().symmetrize()
transformed_traj = transformed.unit_vec_trajectory  # need a better attribute name
```

tests/integration/baseline_images/plot_test/rectilinear.png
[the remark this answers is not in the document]
SCiarella: the old test used to do the transformations ['Normalize', 'Conventional'] before plotting. I have put them back and now the figure is the same.

SCiarella:
Thanks @stefsmeets for the suggestions!

> I'm wondering if we can simplify it a bit, from:
> 
> ```python
> transforms = ['Normalize', 'Conventional', 'Symmetrize']
> vasp_orientations.set_transformation(transforms)
> transformed_traj = vasp_orientations.execute_transformations()
> ```
> 
> to
> 
> ```python
> transforms = ['Normalize', 'Conventional', 'Symmetrize']
> transformed_traj = vasp_orientations.transform(transforms)
> ```
I have used this approach, it makes total sense!

SCiarella:
Moreover, this PR partially covers #286 because now `rectilinear_plot` uses `Orientations.transformed_trajectory` which can be any series of transformations. 
@tfamprikis can you check if this solves the problem described in #286?

src/gemdat/rotations.py
```diff
@@ -24,15 +82,35 @@ class Orientations:
...
     nr_central_atoms: int
-    normalize_trajectory: bool = False
+
+    def __post_init__(self):
+        """Computes trajectories of unit vectors defined as the distance
+        between a central and satellite atoms, meant to track orientation of
+        molecules or clusters."""
+        lattice = self.trajectory.lattice
+        direction = self._fractional_directions(self._distances)
+        self.unit_vec_trajectory = np.matmul(direction, lattice)
```
stefsmeets, line 98: Can you check why these unit vectors do not have length 1?

src/gemdat/rotations.py
```diff
@@ -24,15 +82,35 @@ class Orientations:
...
     center_type: str
     satellite_type: str
     nr_central_atoms: int
-    normalize_trajectory: bool = False
+
+    def __post_init__(self):
+        """Computes trajectories of unit vectors defined as the distance
+        between a central and satellite atoms, meant to track orientation of
+        molecules or clusters."""
+        lattice = self.trajectory.lattice
```
stefsmeets, line 96: I'm a bit confused why this is a numpy array, shouldn't this return a pymatgen `Lattice` object?

```suggestion
        lattice = self.trajectory.get_lattice()
```

src/gemdat/rotations.py
```diff
@@ -247,22 +246,49 @@ def set_symmetry_operations(
...
+            self.transformed_trajectory = self.unit_vec_trajectory
+        else:
+            t_operators = [globals()[name]() for name in transformations]
+
+            # Restart from the original unit vector trajectory
+            self.transformed_trajectory = self.unit_vec_trajectory
+            # then apply all transformations
+            for transformation in t_operators:
+                transformation.transform(self)
+        return self.transformed_trajectory
```
stefsmeets, line 291: Do you think we could return a new `Orientations` class from `transformation.transform`? I think that will fit very nicely with your strategy pattern.

I think it's also better to use a lookup table for the transformation functions to avoid using `globals()` here in this way. That will give some odd error messages in case users mistype the function name.

The lookup table can look something like this:

```
transform_lookup = {
    'normalize': Normalize.transform,
    'conventional': Conventional.transform,
    'symmetrize': Symmetrize.transform,
}
```

The lookup table also shows why we do not necessarily need classes here (see comment above).

```suggestion
        t_operators = [transform_lookup[name] for name in transformations]

        ret = self
        for t_operator in t_operators:
            ret = t_operator(ret)
        
        return ret
```

stefsmeets (commented):
@SCiarella I left some comments in the code!

src/gemdat/rotations.py
```diff
@@ -32,7 +90,31 @@ class Orientations:
...
+        between a central and satellite atoms, meant to track orientation of
+        molecules or clusters."""
+        lattice = self.trajectory.lattice
+        direction = self._fractional_directions(self._distances)
+        self.unit_vec_trajectory = np.matmul(direction, lattice)
+        self.transformed_trajectory = self.unit_vec_trajectory
+        self._prim_to_conv_matrix = np.array(
+            [[1 / np.sqrt(2), -1 / np.sqrt(6), 1 / np.sqrt(3)],
+             [1 / np.sqrt(2), 1 / np.sqrt(6), -1 / np.sqrt(3)],
+             [0, 2 / np.sqrt(6), 1 / np.sqrt(3)]])
```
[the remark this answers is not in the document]
stefsmeets: Pymatgen has `Structure.to_primitive` / `Structure.to_conventional`, which uses `SpacegroupAnalyzer.get_[primitive|conventional]_standard_structure`, which uses `SpacegroupAnalyzer.get_conventional_to_primitive_transformation_matrix`.

PR #303, opened 2024-04-10
Update Orientations api

### LINKED ISSUE #296: Streamline Orientations api
Some ideas to streamline the api:

- [x] `cartesian_to_spherical` -> `Orientations.get_vectors_spherical()`
- [x] `plots.rectilinear_plot(orientations=rotational_traj)` -> `Orientations.plot_rectilinear()` 
- [x] `plots.bond_length_distribution(orientations=rotational_traj, bins=1000)` -> `Orientations.plot_bond_length_distribution()`
- [x] `plots.unit_vector_autocorrelation(orientations=rotational_traj)` -> `Orientations.plot_unit_vector_autocorrelation()`
- [x] `autocorrelation = mean_squared_angular_displacement(traj.vectors)` ->
`Orientations.mean_squared_angular_displacement()`
- [x] `rotations.py` -> `orientations.py`?

Also take note of #301

### LINKED ISSUE #297: Rename 'mean squared angular displacement' to 'autocorrelation' everywhere
In the notebooks and in the `Orientation` class we refer to *mean squared angular displacement*, but in reality what we measure is always the *autocorrelation* (in fact it goes from 1 to 0). 
Correct the name for consistency.

stefsmeets:
This PR adds some convenience methods to Orientations to make the api more accessible.

It also renames `msad` -> `autocorrelation`.

Closes #296 
Closes #297

### Todo

- [x] Update notebook

### DIFF
--- a/src/gemdat/orientations.py
+++ b/src/gemdat/orientations.py
@@ -23,15 +23,12 @@ class Orientations:
         Type of the central atoms
     satellite_type: str
         Type of the satellite atoms
-    nr_central_atoms: int
-        Number of central atoms, which corresponds to the number of cluster molecules
     vectors: np.ndarray
         Vectors representing orientation direction
     """
     trajectory: Trajectory
     center_type: str
     satellite_type: str
-    nr_central_atoms: int
     vectors: np.ndarray = field(init=False)
     in_vectors: InitVar[np.ndarray | None] = None
 
@@ -123,7 +120,11 @@ class Orientations:
         combinations: np.ndarray
             Matrix of combinations between central and satellite atoms
         """
-        index_central_atoms = np.arange(self.nr_central_atoms)
+        nr_central_atoms = frac_coord_cent.shape[1]
+
+        index_central_atoms = np.arange(nr_central_atoms)
+
+        # index_central_atoms = np.arange(self.nr_central_atoms)
         matching_matrix = self._matching_matrix(distance, frac_coord_cent)
         combinations = np.array([(i, j) for i in index_central_atoms
                                  for j in matching_matrix[i, :]])
--- a/src/gemdat/orientations.py
+++ b/src/gemdat/orientations.py
@@ -26,7 +26,7 @@ class Orientations:
     nr_central_atoms: int
         Number of central atoms, which corresponds to the number of cluster molecules
     vectors: np.ndarray
-        Vectors representing rotation direction
+        Vectors representing orientation direction
     """
     trajectory: Trajectory
     center_type: str
--- a/src/gemdat/orientations.py
+++ b/src/gemdat/orientations.py
@@ -6,7 +6,7 @@ import numpy as np
 from pymatgen.symmetry.groups import PointGroup
 
 from gemdat.trajectory import Trajectory
-from gemdat.utils import autocorrelation, cartesian_to_spherical
+from gemdat.utils import fft_autocorrelation, cartesian_to_spherical
 
 
 @dataclass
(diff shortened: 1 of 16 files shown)


stefsmeets:
Hi @SCiarella could you have a look at this?

Particularly:
1. msad -> autocorrelation change
2. if it makes sense to derive `nr_central_atoms` from the trajectory directly

src/gemdat/orientations.py
```diff
@@ -241,6 +240,40 @@ def transform(self, matrix: np.ndarray) -> Orientations:
...
+        ----------
+        degrees : bool
+            If True, return angles as degrees, else radians
+
+        Returns
+        -------
+        np.array
+            azimuth, elevation, length
+        """
+        return cartesian_to_spherical(self.vectors)
```
SCiarella, line 256: Maybe this can also return an `Orientation` doing something like
```suggestion
        return replace(self, in_vectors=cartesian_to_spherical(self.vectors))
```
similarly to `normalize`, `transform` and `symmetrized`.

src/gemdat/utils.py
```diff
@@ -229,32 +229,74 @@ def _cart2sph(x: np.ndarray, y: np.ndarray,
...
+    Parameters
+    ----------
+    coords : np.ndarray
+        The input signal in direct cartesian coordinates. It is expected
+        to have shape (n_times, n_particles, n_coordinates)
+
+    Returns
+    -------
+    autocorrelation: np.array
+        The autocorrelation of the input signal
```
SCiarella, line 276:
```suggestion
        The autocorrelation of the input signal, with shape (n_particles, n_times)
```

src/gemdat/utils.py
```diff
@@ -229,32 +229,74 @@ def _cart2sph(x: np.ndarray, y: np.ndarray,
...
+        # Compute the inverse FFT of the power spectral density
+        autocorr_c = np.fft.irfft(fft_signal, axis=0)
+
+        # Only keep the positive times
+        autocorr_c = autocorr_c[:n_times, :]
+
+        autocorrelation += autocorr_c.T / normalization
+
+    # Normalize the autocorrelation such that it starts from 1
+    # (this makes the normalization independent on the dimensions)
```
SCiarella, line 299:
```suggestion
    # and make it independent of n_coordinates
```

tests/orientations_test.py
```diff
@@ -5,30 +5,27 @@
...
     orientations = Orientations(trajectory=trajectory,
                                 center_type='B',
-                                satellite_type='Si',
-                                nr_central_atoms=1)
+                                satellite_type='Si')
 
     assert isinstance(orientations, Orientations)
     assert orientations.center_type == 'B'
     assert orientations.satellite_type == 'Si'
-    assert orientations.nr_central_atoms == 1
```
SCiarella, line 23: something like
```suggestion
    assert orientations.trajectory == trajectory
```

SCiarella (changes requested):
Looks awesome! 

I added a few comments. In particular, we could think about making `cartesian_to_spherical` return an `Orientation` and behave like the other transformations.

----
> 1. msad -> autocorrelation change

Looks perfect

> 2. if it makes sense to derive nr_central_atoms from the trajectory directly

Yes, it makes total sense

src/gemdat/orientations.py
stefsmeets: Maybe it makes more sense to have this as a property:

```Orientations.vectors_spherical```

stefsmeets: Ah but that loses the degrees flag 🤔

stefsmeets: The degrees flag was not passed, so that seems like a bug anyway.

Issue #306, opened 2024-05-02
Return molecule trajectory from Orientations


stefsmeets:
We already have all the information for the grouping of molecules including their orientations. We could use those for a method that returns a molecule trajectory (essentially a `Trajectory` without `Lattice`).

PR #307, opened 2024-05-02
Orientational sites

SCiarella:
This PR uses [blob_dog](https://scikit-image.org/docs/stable/api/skimage.feature.html#skimage.feature.blob_dog) to identify the orientations which are explored more often.

Now running `plot_rectilinear(add_peaks=True)` you can obtain figures like this one:
a

The calculation of the peaks has been implemented by introducing an `OrientationalVolume` class, inherited from `Volume`.

------

This PR may be redundant once #306 is implemented


src/gemdat/volume.py
```diff
@@ -442,6 +444,51 @@ def optimal_percolating_path(self, **kwargs) -> Pathway | None:
...
+        -------
+        peaks : list
+            List of coordinates of the peaks
+        """
+        kwargs.setdefault('threshold', 0.1)
+
+        theta, phi, values = self.data.T
+
+        x = theta
+        y = phi
```
stefsmeets, line 481:
```suggestion
        theta, phi, values = self.data.T
```

stefsmeets (commented):
Just some comments on this PR. I'm not entirely sure why we need to have a subclass of Volume for what is essentially peakfinding a 2D histogram. It's a lot to go through to understand what is going on, especially because in the plots, the code also goes through the OrientationalVolume. Conceptually, I think it would be easier to understand this if all the methods were self-contained on Orientations. This is mostly about rearranging bits, other than that it looks pretty good 😅

stefsmeets:
I fixed all the merge conflicts with main in this PR

stefsmeets:
I'm going to close this for now because of inactivity and no clear path forward.

PR #323, opened 2024-05-31
Add warning for multiple paths

### LINKED ISSUE #275: Path enhancements
- [ ] `Pathway` objects should return their total length in Angstrom
- [ ] `Pathway` should also compute and report the energy difference between connected sites, and report this information which is more relevant than the total free energy of the sites
- [ ] The current implementation for `multiple_paths` is inefficient if the user wants paths with large differences. The reason is that the algorithm is based on `netwokx.all_shortest_paths` which starts the search from the optimal one with small variations. We should address that:
- - [ ] the documentation of `gemdat.multiple_paths` should report that it is suited only for small path variations, 
- - [ ] an approach with graph pruning may be optimal to identify strongly distinct paths, but this may deserve a separate issue
- - [ ] the shaded background in the `path_energy` plots representing the sites explored, should be made optional in case of multiple paths that explore different sites

SCiarella:
This PR closes #275 by addressing task 3.1.

### DIFF
--- a/src/gemdat/path.py
+++ b/src/gemdat/path.py
@@ -298,6 +298,11 @@ def optimal_n_paths(
     (https://doi.org/10.1287/mnsc.17.11.712) and its implementation in NetworkX.
     Only paths that are different by at least min_diff are considered.
 
+    .. warning::
+        Notice that this function in based on networkx.all_shortest_paths, which tends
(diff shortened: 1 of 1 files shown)


stefsmeets (approved):
Nice 'fix' 😅

Issue #16, opened 2023-06-27
Calculate sites data from materials input


stefsmeets:
This issue tracks processing the 'known materials' data and calculating sites data.

### Features

- [ ] Read known materials from standard crystallographic input (wyckhof symbol, cif file, or otherwise?)

### Variables

- [ ] occupancy
- [ ] occupancy_parts
- [ ] atoms
- [ ] transitions
- [ ] all_transitions
- [ ] succes
- [ ] sites_occupancy
- [ ] atom_locations
- [ ] atom_locations_parts
- [ ] jump_names
- [ ] nr_jumps
- [ ] rates
- [ ] e_act
- [ ] collective
- [ ] collective_jumps
- [ ] collective_matrix
- [ ] multi_collective
- [ ] solo_jumps
- [ ] solo_frac
- [ ] jump_diffusivity
- [ ] correlation_factor

PR #324, opened 2024-06-01
Add plotly plot for shape

### LINKED ISSUE #315: Add plotly plot for shape analysis
After #313 only the plot for shape analysis is still missing.

See also #302

SCiarella:
This PR closes #315 implementing the shape plot using plotly, like this:
aaa

Testing is also implemented.

Notice that I have also set the requirement for `pytest>=8.1` because of some rounding errors that make the plotly test fail. 

----

Before merging this PR, it is still needed to check if 
- [ ] the new plotly plot works for more complex configurations 
- [ ] it accommodates all the operation performed in [this example](https://nbviewer.org/github/GEMDAT-repos/gemdat-notebooks/blob/main/shape_analysis.ipynb)

This can be fulfilled following [this link](https://github.com/GEMDAT-repos/gemdat-notebooks/pull/16) and testing the notebook with the correct input data.


src/gemdat/plots/plotly/_shape.py
```diff
@@ -32,4 +34,131 @@ def shape(
...
+            col=col + 1,
+        )
+
+        # Highlight the (0,0,0) point
+        x0 = [0]
+        y0 = [0]
+        fig.add_trace(
+            go.Scatter(
+                x=x0,
+                y=y0,
```
stefsmeets, line 91:
```suggestion
        fig.add_trace(
            go.Scatter(
                x=[0],
                y=[0],
```

src/gemdat/plots/plotly/_shape.py
```diff
@@ -32,4 +34,131 @@ def shape(
...
+            go.Histogram2dContour(
+                x=x_coords,
+                y=y_coords,
+                name=y_labels[col],
+                showscale=False,
+                colorscale='Viridis',
+            ),
+            row=1,
+            col=col + 1,
+        )
```
stefsmeets, line 83: Might be nice to only define the row/col once.

```suggestion
        top_row = {'row': 1, 'col': col + 1}
        bot_row = {'row': 2, 'col': col + 1}

        # Map of the coordinates
        fig.add_trace(
            go.Histogram2dContour(
                x=x_coords,
                y=y_coords,
                name=y_labels[col],
                showscale=False,
                colorscale='Viridis',
            ),
            **top_row
        )
```

src/gemdat/plots/plotly/_shape.py
```diff
@@ -32,4 +34,131 @@ def shape(
...
+        y_coords = coords[:, j]
+
+        # Map of the coordinates
+        fig.add_trace(
+            go.Histogram2dContour(
+                x=x_coords,
+                y=y_coords,
+                name=y_labels[col],
+                showscale=False,
+                colorscale='Viridis',
```
stefsmeets, line 79: I think we should not define the colormap ourselves, and just use the default for consistency with the other plots.

```suggestion
```

src/gemdat/plots/plotly/_shape.py
```diff
@@ -32,4 +34,131 @@ def shape(
...
+            # Add shape points
+            fig.add_trace(
+                go.Scatter(
+                    x=x_vs,
+                    y=y_vs,
+                    mode='markers+text',
+                    text=[label] * len(x_vs),
+                    textposition='top center',
+                    textfont=dict(color='red'),
+                    marker=dict(color='red'),
```
stefsmeets, line 128: Cleaner and faster to use a dict literal, imo.

```suggestion
                    textfont={'color':'red'},
                    marker={'color':'red'},
```

src/gemdat/plots/plotly/_shape.py
```diff
@@ -32,4 +34,131 @@ def shape(
...
+            # Area around the shape points
+            fig.add_shape(
+                type='circle',
+                xref='x',
+                yref='y',
+                x0=x_vs - mean_dist,
+                y0=y_vs - mean_dist,
+                x1=x_vs + mean_dist,
+                y1=x_vs + mean_dist,
+                line=dict(color='red', dash='dash'),
```
stefsmeets, line 144:
```suggestion
                line={'color':'red', 'dash':'dash'),
```

src/gemdat/plots/plotly/_shape.py
```diff
@@ -32,4 +34,131 @@ def shape(
...
+    vector_dict = {}
+    if sites:
+        for site in sites:
+            _tmp_dict[site.label].append(site.coords)
+        for key, values in _tmp_dict.items():
+            vector_dict[key] = np.array(values) - shape.origin
+
+    coords = shape.coords
+
+    for col, (i, j) in enumerate(((0, 1), (1, 2), (2, 0))):
```
stefsmeets, line 68: Oof, this for loop is a bit of a beast 😅 I don't see an obvious way to make it better though.

Perhaps you can factor out the inner and outer loops like so, although that requires passing around a ton of variables:

```
    for col, (i, j) in enumerate(((0, 1), (1, 2), (2, 0))):
        _plot_contour(fig=fig, loc=top_row, coords=coords, ...)

        for label, vects in vector_dict.items():
            _plot_histo(fig=fig, loc=bot_row, vects=vects, ...)
```

src/gemdat/plots/plotly/_shape.py
```diff
@@ -21,9 +24,8 @@ def shape(
     ----------
     shape : ShapeData
         Shape data to plot
-    bins : int | Sequence[float]
-        Number of bins or sequence of bin edges.
-        See [hist()][matplotlib.pyplot.hist] for more info.
+    nbins : int
+        Number of bins
```
stefsmeets, line 28: With this change the api is no longer consistent between the matplotlib version and the plotly version.

stefsmeets (commented):
Hi @SCiarella , nice work! This plot looks great! I have some nitpicks in the comments, see if you find them useful to do something with. 

I have 2 main comments:
- The api is no longer consistent between with the matplotlib version. Could you make sure that they both have the same keyword arguments? Either bins or nbins is fine for me.
- The plots are missing labels for the y axes.

src/gemdat/plots/plotly/_shape.py
SCiarella: done

src/gemdat/plots/plotly/_shape.py
SCiarella: True! I made them consistent using `bins`

src/gemdat/plots/plotly/_shape.py
SCiarella: fixed

SCiarella:
I have implemented the suggested changes, in particular:
* now the API is consistent using `bins` in both cases
* the axis has been fixed

I have not improved however the loop because I can not identify an obvious solution either. 

Before finalizing this PR, I would like to test using the same data used [here](https://nbviewer.org/github/GEMDAT-repos/gemdat-notebooks/blob/main/shape_analysis.ipynb). @stefsmeets Can you share them?

stefsmeets:
Nice work! 🚀

stefsmeets (approved):

stefsmeets:
Now we have all the plots for both matplotlib and plotly 😁

Issue #47, opened 2023-07-11
Add support for data from other formats


stefsmeets:
See available formats here:

https://pymatgen.org/pymatgen.io.html

v1kko:
I looked into this a bit, and I find that probably the best solution is something very close to the one that we already have.
The `Data` structure can be created with its arrays through the default constructor which can then be cached by the user if desired. This would generate the following code for any data format:

```python
from pymatgem.io import correct_parser
from gemdat import Data

def generic_reader(file, cache):
  if Path(cache).exists():
          return Data.from_cache(cache)
  some_object = correct_parser(file);
  # Massage `some_object` into correct format for `Data`
  massaged_data = {
              'structure'=structure, 
              'trajectory_coords': trajectory.coords,
              'species': structure.species,
              'lattice': structure.lattice,
              'time_step': run.parameters['POTIM'] * 1e-15,
              'temperature': run.parameters['TEBEG'],
              'parameters': run.parameters, }
  return Data(**massaged_data)

generic_reader("file","cache")
data.to_cache("cache")
```
I think we should offer this function as a way people can add new readers, for new data formats. If they have created one we can then implement it in `GEMDAT`.

It might be possible to modularize the `from_xxx` functionality, but that would come at the cost of increased complexity

v1kko:
@alexandrosvasi @AstyLavrinenko Please provide us with example data for Lamps/gromacs/cp2k

stefsmeets:
@alexandrosvasi @AstyLavrinenko Can you give us an update on this?

stefsmeets:
We now have some data to test with for LAMMPS

PR #330, opened 2024-07-11
Add GROMACS support

AstyLavrinenko:
#47 
I uploaded gromacs data to test_data repository. 

I uploaded here the notebook test_gromacs_support.txt where I tested function from_gromacs. It seems working fine, but I have some concerns:

- I didn't add cache support to the function
- It's a bit weird when extracting structure. Structure information starts from some line not lattice (see the notebook). But lattice works.
- I'm using site_properties attribute of Trajectory to introduce residues
- It could be cool to have opportunity to filter trajectory by residue (see the notebook)
- I got out_of_memory problem when plotting msd from large simulation

[msd.csv](https://github.com/user-attachments/files/16195323/msd.csv)
[test_gromacs_support.txt](https://github.com/user-attachments/files/16195325/test_gromacs_support.txt)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,7 +24,7 @@ repos:
     hooks:
       - id: docformatter
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.4.4
+    rev: v0.5.2
     hooks:
       - id: ruff
         args: [--fix]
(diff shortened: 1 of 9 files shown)


stefsmeets:
Hi @AstyLavrinenko thanks for this. I fixed a few small issues with the code. The plot labels were indeed a bug that incorrectly assumed that trajectory stores the time_step in seconds instead of fs.

The out of memory issue is somewhat to be expected with such large systems, especially with lots of time steps.

I will add some tests and then this can be merged.

stefsmeets (approved):
Hi @AstyLavrinenko I'm going to merge this now, please let me know if you run into any issues.

PR #336, opened 2024-10-14
Fix vibration amplitudes plot

### LINKED ISSUE #335: Plot vibrational amplitudes
The plot of vibrational amplitudes yields different results between matplotlib and plotly backends, with matplotlib showing occurrences up to 2.5 and plotly showing up to 2.0 and appearing more symmetric for some reason.

image
image

Additionally, matplotlib doesn't support `bins` and `n_parts` keywords, which would be beneficial since matplotlib often performs faster and is more convenient in some cases.

stefsmeets:
This PR updates the vibratitonal amplitudes plot so that the matplotlib version matches the plotly version.

Closes #335

### Todo
- [x] Fix mypy issue with matplotlib.figure.Figure
- [x] Add n_parts support
- [x] Plot error bars on mpl function
- [x] Check consistency of all other plotly/mpl plots


stefsmeets:
Plots are now generated in exactly the same way, and should show the same histogram:

PR #338, opened 2024-10-16
Add RDFs calculation based on species

AstyLavrinenko:
Currently, the RDFs can be calculated for floating species, requiring sites data or transition information. Could you please additionally implement a version that calculates RDFs between specified species without the need for transition data. I have provided the code we use for this calculation, which includes parallelization.


stefsmeets:
Hi @AstyLavrinenko thanks for this, I refactored the code a bit. I think I'm going to add it as a plot function to Trajectory:

```
trajectory.plot_rdf_between_species(...)
```

Would that work for you?

### Todo
- [ ] Add method on trajectory
- [ ] Move hist code to plots/matplotlib
- [ ] Add plotly version
- [ ] Add integration test

stefsmeets:
Let me know if you are happy with this!

AstyLavrinenko:
Hi @stefsmeets thanks a lot!
The only concern is that raw data is not easily accessible now, and it's nice to have such opportunity as sometimes we need to visualize several RDFs at one plot. Can we maybe move `_get_radial_distribution_between_species` function as a `Trajectory` function?

stefsmeets:
Yeah, makes sense. I came up with a better way to organize the code, I will make a PR soon.

PR #342, opened 2024-11-18
Fix crash in `Transitions.occupancy` if input sites are disordered

### LINKED ISSUE #339: Problem with occupancy function
I'm trying to get working functions `transitions.occupancy()` and `transitions.atom_locations()` and getting an error:
image

After fixing `specie` to `species` in the source code, I'm still getting error:
image

Could you please fix that?

Besides that occupancy now seems to appear in the `Structure` object. Is there any additional function to extract average occupancy per each site type?

stefsmeets:
This PR adds a  work-around when `Transitions.occupancy` is called, but the input sites are disordered. Disordered jump sites technically do not make sense, but the code only works with the coordinates anyway.

Closes #339

### DIFF
--- a/src/gemdat/transitions.py
+++ b/src/gemdat/transitions.py
@@ -275,7 +275,22 @@ class Transitions:
             labels=sites.labels,
         )
 
-    def atom_locations(self):
+    def occupancy_by_site_type(self) -> dict[str, float]:
+        """Calculate average occupancy per a type of site.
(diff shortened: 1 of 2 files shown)


AstyLavrinenko:
Hi @stefsmeets 
I just commented on the issue, but we also can have a quick call if more details are needed. Please let me know

PR #347, opened 2025-01-06
Update the Colormap for Shape Analysis

AstyLavrinenko:
Hello Stef,

I have updated the colormap for the shape analysis with `matplotlib` backend to match the colormap used in the density plot. Could you please incorporate them into the `plotly` backend as well? 
I have attached the output for the reference. 
image

Thank you in advance!


stefsmeets:
Thanks for the PR. I want to stick as close to the matplotlib/plotly defaults where possible, because it is so dependent on personal preference. If we use any non-default colors, I consider that a bug. I usually try to avoid hard-coding the color map in the plotting code, and rather pass a color map separately. Do you think you could make this change?

AstyLavrinenko:
Hi @stefsmeets ,
Don't we have this colormap in density plot? Is it possible to use same here?

stefsmeets:
Hi @AstyLavrinenko ,  updated the api so that it takes a colormap as an argument. Hope this works for you. If we need to have a custom colormap for this function, perhaps we can have a `gemdat.colormaps` module where we store those.

AstyLavrinenko:
> Hi @AstyLavrinenko , updated the api so that it takes a colormap as an argument. Hope this works for you. If we need to have a custom colormap for this function, perhaps we can have a `gemdat.colormaps` module where we store those.

Hi @stefsmeets it would be nice to have this colormap coded especially because it is supposed to be used for both density plot and shape analysis. Could you please set this colormap as a default version for the shape analysis plot?

stefsmeets:
> > Hi @AstyLavrinenko , updated the api so that it takes a colormap as an argument. Hope this works for you. If we need to have a custom colormap for this function, perhaps we can have a `gemdat.colormaps` module where we store those.
> 
> Hi @stefsmeets it would be nice to have this colormap coded especially because it is supposed to be used for both density plot and shape analysis. Could you please set this colormap as a default version for the shape analysis plot?

I'm against using rainbow colormaps (e.g. 'jet') as a sequential colormap, because they are misleading for continous data. I have revised the colormap for the density plot to make it more inline with the standard plotly colormap in #359. 

You can still easily get the rainbow colormap in gemdat for the shape data:

```python
shape.plot(bins=bins, backend='matplotlib', cmap='jet')
```

PR #359, opened 2025-04-01
Add cmap option to plotly shape function

stefsmeets:
Add cmap option to plotly plot.


AstyLavrinenko:
Hi @stefsmeets! I spoke with Alex and we would like to return to the previous color palette, as it offers the higher contrast needed for the density plot. Could you please switch both the 3D density plot and the shape analysis plot back to that previous color palette? Additionally, could you please add a color scale bar to both plots?

stefsmeets:
> Additionally, could you please add a color scale bar to both plots?

It is not a continuous color, so a colorbar does not make much sense here. I have reverted the colors.

PR #362, opened 2025-04-03
Add flag to set all occupancies to 1

### LINKED ISSUE #343: Dealing with partial occupancies
This applies to analyses that require sites/user structure. If a user provides a disordered structure, the code should return an error indicating that the analysis doesn't work with partial occupancies. The user may then either manually adjust the structure or, if they prefer an automatic solution, set flag = True. If flag = True is enabled, the code should issue a warning that the structure will be modified: all partial occupancies will be reset to 1, and any issues with duplicate sites will be addressed. 

Forcing all occupancies to 1 may introduce problems if the structure contains elements other than the floating species.
In such cases, we have two options:
- remove all sites containing other species
- if available, use a function to remove duplicated species.

I'm not sure that pymatgen has a function to remove duplicated species, therefore, for my structures, I used `custom_structure.remove_species(['S','Sb','W'])` to delete other species and the following to change the occupancy to 1:
```
for idx,site in enumerate(custom_structure.sites):
    custom_structure.replace(idx=idx, species=site.species.elements[0], label=site.label)
```

stefsmeets:
Closes #343

### DIFF
--- a/src/gemdat/transitions.py
+++ b/src/gemdat/transitions.py
@@ -49,6 +49,7 @@ class Transitions:
         events: pd.DataFrame,
         states: np.ndarray,
         inner_states: np.ndarray,
+        set_partial_occupancies_to_1: bool = False,
     ):
         """Store event data for jumps and transitions between sites.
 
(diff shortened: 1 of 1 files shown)


stefsmeets:
@AstyLavrinenko Let me know if this works for you.

PR #364, opened 2025-04-08
Raise error when disordered structure is passed to `Transitions`

### LINKED ISSUE #343: Dealing with partial occupancies

stefsmeets:
This PR changes how disordered species are handled by `Transitions`. It now raises an error, because the analysis does not work with disordered species. I also added this `Transitions.from_trajectory`), so it now breaks at the earliest possible stage. I added a function `gemdat.utils.remove_disorder_from_species` to help 'fix' the structure. 

This fixes 2 issues with the current code:

1. The current code updates `sites` in-place. This can give unintended side effects
2. There is better separation of concerns. This changes allows the user to inspect the `sites` before using the `Transitions` code. This prevents unwanted changes to the structure and places control for making sure the structure is what they wanted with the users.

Usage:

```python
from gemdat.utils import remove_disorder_from_structure

new_structure = remove_disorder_from_structure(structure)
```

Closes #343

### DIFF
--- a/src/gemdat/trajectory.py
+++ b/src/gemdat/trajectory.py
@@ -33,7 +33,7 @@ SP_NAME = re.compile(r'([a-zA-Z]+)')
 
 
 def _lengths(vectors: np.ndarray, lattice: Lattice) -> np.ndarray:
-    """Calculate vector lengths using the metric tensor (Dunitz 1078, p227).
+    """Calculate vector lengths using the metric tensor (Dunitz 1978, p227).
 
     Parameters
     ----------
--- a/src/gemdat/transitions.py
+++ b/src/gemdat/transitions.py
@@ -42,6 +42,7 @@ class Transitions:
 
     DISORDER_ERROR_MSG = (
         'Input `sites` are disordered! '
+        'The analysis does not work with disordered structures. '
         'Remove disorder and partial occupancies or try '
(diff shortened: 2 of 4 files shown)


AstyLavrinenko:
Hi @stefsmeets this function still doesn't address the problem with duplicated sites after setting the occupancies to 1. Could you please fix this.

Also for the error could please change the text from "disorder" to "partial occupancy" as disorder implies ordered and disordered phases while we state here only the existence of all possible for occupation sites which is not necessarily ordered phase.

PR #19, opened 2023-06-27
Add implementation for calculating site transitions

stefsmeets:
This PR adds an implementation to calculate site transitions.

This code covers everything until line 156 in `find_sites.m`.

Data are loaded from a cif file, which means they cannot be compared directly to the matlab data (the order of the sites and thus coordinates is different).

The result differs slightly, I think because of rounding errors.

### Todo
- [x] Generate `succes` array
- [x] Generate `transitions` array
- [x] Generate `occupancy` array
- [x] Generate `occupancy_parts` array
- [x] Generate `atom_sites` array
- [x] Refine and refactor code
- [x] Load `vibration_amplitude` from some sort of cache or calculate it on-the-fly (Maybe a `preprocess(Data)` to get everything at once?) -> #20

PR #403, opened 2026-06-09
Goac integration

SCiarella:
This PR adds support for [GOAC](https://iffgit.fz-juelich.de/k.koester/goac/-/tree/master/) as an optional dependency, including an automated installation script, tests, CI, and docs.

In particular I have added:
* `install_goac.py`: helper script that checks system dependencies (gfortran, meson, ninja, OpenMP) and Python deps, then compiles and installs GOAC via f2py. 
* `install_goac_test.py` tests for the install script.
* `goac.yaml` CI workflow that installs system dependencies and runs the GOAC tests on PRs/pushes to main. For now it is not part of `test.yaml` because I did not want to force the goac dependencies to the test GHrunner.
* fixes to `README.md` & `CONTRIBUTING.md` adding installation instructions for GOAC .

There is also a [new notebook](https://github.com/GEMDAT-repos/gemdat-notebooks/pull/19) to show how to use GOAC.


tools/install_goac.py
```diff
@@ -0,0 +1,233 @@
+#!/usr/bin/env python3
+"""Install GOAC (Global Optimization of Atomistic Configurations by Coulomb).
```
v1kko: I think we should create a new repository (GEMDAT-repos/GOAC) in which we build GOAC as a package and provide installable wheels as an artifact for windows/mac/linux.

I don't think we should put it on pypi just yet, that should be done in collaboration with the original author

v1kko: So we shouldn't include this file here

tests/install_goac_test.py
```diff
@@ -0,0 +1,47 @@
+"""Test that the GOAC installation helper script works end-to-end."""
```
v1kko: Same, shouldn't be included in this PR imo

docs/notebooks
v1kko, line 1: I dont think we should update the notebooks submodule to another branch then main. The goac notebook is now in main, but we don't refer to it from the actual documentation anyway...

.gitignore
```diff
@@ -122,6 +122,8 @@ pyvenv.cfg
 
 # vscode
 .vscode/
+.github/copilot-instructions.md
```
v1kko, line 125: As a thought, can we improve findability of the repository by explicitly adding these kind of files (also CLAUDE.md)?

CONTRIBUTING.md
```diff
@@ -56,6 +58,12 @@ source env/bin/activate
 python3 -m pip install -e .[develop]
 ```
 
+For optional GOAC support (requires gfortran, meson >= 1.6.0, ninja >= 1.11.1.2, OpenMP):
+```console
+sudo apt-get install gfortran meson ninja-build libomp-dev  # Ubuntu/Debian
+python3 tools/install_goac.py
+```
+
```
v1kko: Id like to update this to 
```console
pip install GOAC \
    --no-index \
    --find-links https://github.com/GEMDAT-repos/GOAC/releases/expanded_assets/vx.x.x
```
so we can cleanly update it later to a pypi package

README.md
```diff
@@ -28,6 +28,42 @@ To install:
 pip install gemdat
 ```
 
+### Optional Dependencies
```
v1kko, line 31: again, this can be a simple pip install from GEMDAT-repos/GOAC

.github/workflows/tests.yaml
```diff
@@ -50,7 +50,7 @@ jobs:
 
       - name: Test with pytest
         run: |
-          coverage run -m pytest
+          coverage run -m pytest --ignore=tests/install_goac_test.py
```
v1kko:
```suggestion
          coverage run -m pytest
```
I'd remove the install_goac_test.py and move it to the GOAC repo

v1kko (commented):
Interesting work, but I am not sure if GEMDAT is the right place to put the GOAC installers. I'd much prefer them to live in e.g. GEMDAT-repos/GOAC temporarily until we make an installable package with the GOAC maintainers.

docs/notebooks
SCiarella: ah yes, I forgot about that. But I had in mind to display it like the others in the docs

.gitignore
SCiarella: mmm maybe, but those files can contain rules such as *respond to me in italian* that I don't think are useful for every user

SCiarella:
Thanks @v1kko!

> Interesting work, but I am not sure if GEMDAT is the right place to put the GOAC installers. I'd much prefer them to live in e.g. GEMDAT-repos/GOAC temporarily until we make an installable package with the GOAC maintainers.

I like this (temporary) solution. I will open a temporary `GEMDAT-repos/GOAC` and make the install cleaner without the need for the tool

SCiarella:
@v1kko there is now an error with python3.14 that is not able to pixel match the baseline images anymore. Any idea why this failure happens only now?

v1kko:
Probably dependencies are updated, deps for 3.11 and 3.12 seem cached, we can get the package list for 3.14 from "install dependencies", but not for 3.11 and 3.12 to scan for differences as those are cached ...

v1kko:
If the tests work on your machine, could you get a package set? (pip freeze or similar)

SCiarella:
actually it does not work anymore on my machine. Not sure at which point it happened, but i think that baseline images needs to be updated

v1kko:
Or we pin the package that caused the change, it should be reproducible after all...
Let me try to extract the versions for the 3.11 and 3.12 pipeline

v1kko:
the tests are fixed

Issue #406, opened 2026-07-07
drift(floating_species=...) crashes on trajectories with Element species (e.g. from_vasprun)


v1kko:
### Description

`Trajectory.drift()` (and therefore `apply_drift_correction()`) crashes when called with `floating_species` on a trajectory whose species are `pymatgen` `Element` objects rather than `Species`.

In `trajectory.py`, the `floating_species` branch asserts every site is a `Species`:

```python
elif floating_species:
    species = set()
    for sp in self.species:
        assert isinstance(sp, Species), f'got {type(sp)=}'   # <-- fails for Element
        if sp.symbol not in floating_species:
            species.add(sp)
```

Trajectories loaded via `Trajectory.from_vasprun(...)` have `Element` species, so this assertion fails:

```
AssertionError: got type(sp)=<enum 'Element'>
```

The `fixed_species` branch does not have this problem, so `drift(fixed_species=...)` works while `drift(floating_species=...)` does not.

### Steps to reproduce

```python
from gemdat import Trajectory

traj = Trajectory.from_vasprun('vasprun.xml')
traj.apply_drift_correction(floating_species='Li')   # AssertionError
```

### Expected behaviour

`drift`/`apply_drift_correction` should accept both `Element` and `Species` sites (both expose `.symbol`), consistent with the rest of the API (e.g. `filter()`).

### Notes

Found while fixing #394 (variable-lattice handling); this bug is independent of the lattice issue.

PR #407, opened 2026-07-07
Fix lattice handling when constant_lattice=False (#394)

### LINKED ISSUE #394: Issues with lattice when constant_lattice = False
* Several functions call `self.get_lattice()` (for example, `Trajectory.filter()`, `Trajectory.mean_squared_displacement()`, ...). But in case of `constant_lattice = False` `self.get_lattice()` requires `idx` and doesn't work without it causing crash of `Trajectory.filter()`, `Trajectory.mean_squared_displacement()`, ... .

* `Trajectory.get_lattice()` also calls `latt = self.lattices[idx]`, but `self` doesn't have an attribute `lattices`.

* Function `Trajectory.from_gromacs()` creates variable `lattices` if not `constant_lattice` but in the output object uses `lattice` anyway.

v1kko:
Fixes #394.

Several `Trajectory` methods broke for a variable (NPT) lattice loaded with `constant_lattice=False`. This PR fixes them and adds regression tests.

## Changes

- **`get_lattice()`** with no index previously did `self.lattice[None]` (the whole per-frame array), crashing on `Lattice(...)` reshape. It now **raises a clear `ValueError`** when the lattice is not constant and no frame index is given.
- **`mean_squared_displacement()`** and **`distances_from_base_position()`** called `get_lattice()` with no index. They now convert **each frame to cartesian using that frame's own lattice** — the correct treatment for a varying cell.
- **`center_of_mass()`** and **`apply_drift_correction()`** passed `lattice=self.get_lattice()` and dropped `constant_lattice`, losing the per-frame lattice. They now forward `self.lattice` and `constant_lattice`, like `filter()` already does.
- **`to_ase_trajectory()`** used `get_lattice()` for the non-constant branch, which no longer yields the per-frame array; it now reads `self.lattice` directly.
- **`from_gromacs()`** built a `lattices` list for `constant_lattice=False` but passed the undefined `lattice` to the constructor; it now assigns it.

## Tests

Adds regression tests covering `get_lattice`/`filter`/`mean_squared_displacement` on a variable lattice, including that the per-frame MSD path matches the constant-lattice path when all cells are identical. Full unit + integration suites pass; pre-commit (ruff, ruff-format, docformatter, mypy) is clean.

## Note

While verifying on real VASP data I found an unrelated pre-existing bug in `drift(floating_species=...)` — filed as #406. It is out of scope for this PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/gemdat/trajectory.py
AstyLavrinenko, line 1: I noticed that for method `from_lammps` we have no support in case of not constant lattice. Is it possible to fix this and add it?

AstyLavrinenko (changes requested):
There is still a problem in the integration with other functions calculating properties from a trajectory. It would be helpful to fix `particle_density` in `metrics.py` to work with a variable lattice and calculate the volume per lattice.

For the functions below, a not-constant-lattice version is not physically meaningful, so it would be better to raise a not-supported error, since the error in `trajectory.get_lattice()` will otherwise just ask to add a frame index:
- `from_given_radius`, `from_vibration_amplitude`, and `_calculate_atom_states` in `transitions.py`
- `jump_diffusivity` and `collective` in `jumps.py`
- `_compute()` in `collective.py`
- `radial_distribution` in `rdf.py` (`radial_distribution_between_species` works with a variable lattice, which is fine because it isn't based on `Transitions`)
- `framework` in `crystallizer.py`
- `analyze_trajectory` in `shape.py` only checks similarity with `trajectory.get_lattice(0)` and warns if the lattices are not close. For a not-constant lattice, it should raise a not-supported error.

The following could probably be physically meaningful with a variable lattice, but I don't think it's useful to fix now, so please add a not-supported error too:
- `trajectory_to_volume` in `volume.py`
- `__post_init__` and `_distances` in `orientations.py`

src/gemdat/trajectory.py
v1kko: Yes, but it would require some work, so I created a separate issue #417

src/gemdat/orientations.py
```diff
@@ -52,9 +52,15 @@ def __post_init__(self, in_vectors: np.ndarray | None = None):
...
             direction = self._fractional_directions(self._distances)
-            self.vectors = lattice.get_cartesian_coords(direction)
+
+            if self.trajectory.constant_lattice:
+                lattice = self.trajectory.get_lattice()
+                self.vectors = lattice.get_cartesian_coords(direction)
+            else:
+                # Convert each frame with its own lattice (e.g. NPT).
+                matrices = np.asarray(self.trajectory.lattice, dtype=float)
+                self.vectors = np.einsum('fai,fik->fak', direction, matrices)
```
v1kko, line 63: It is possible to make orientations work with a variable lattice, as the vectors can "compensate" for the lattice deformation. So instead of a NotImplementedError all orientation functions are implemented

src/gemdat/shape.py
```diff
@@ -231,6 +231,7 @@ def find_equivalent_positions(
         cart_coords = self.lattice.get_cartesian_coords(centered)
         return cart_coords
 
+    @require_constant_lattice
     def analyze_trajectory(
```
v1kko, line 235: This function used to work (incorrectly), now it throws an error

v1kko (commented):
I think this guards all non-implemented cased. If we want they can be implemented separately later

